PR #5, opened 2023-11-08
Add mf arch config files

dareg:
Add the config files for Meteo-France cluster in the same way ECMWF does it.


awnawab (approved):
Thanks for adding this @dareg! Looks great imo, g2g.

Issue #4, opened 2023-11-08
Separate CPU and GPU tests


awnawab:
Currently, the CI runs on CPU using GNU compilers with OpenACC disabled. In this case, OpenACC directives are treated as comments. The `*_gpu` tests therefore are not run in their intended configuration in the CI and passing tests might give one a false sense of reassurance.

I propose we split the tests into a CPU and GPU list, and the GPU tests are only run if OpenACC is enabled. This would mean that until the github CI can be configured to run on GPU, we would have to test the `*_gpu` tests manually.

PR #9, opened 2023-11-13
Separate GPU specific tests

awnawab:
This PR implements the enhancement highlighted in #4. In a nutshell, GPU specific tests are only built if OpenACC is found.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -124,3 +124,16 @@ ecbuild_add_test(
     LINKER_LANGUAGE Fortran
     CONDITION ${HAVE_SINGLE_PRECISION}
 )
+
+## Test presence of GPUs
+add_executable(check_gpu_num.x check_gpu_num.F90)
+target_link_libraries(check_gpu_num.x PRIVATE $<${HAVE_ACC}:OpenACC::OpenACC_Fortran>)
+set_target_properties(check_gpu_num.x PROPERTIES LINKER_LANGUAGE Fortran)
+
+if(HAVE_ACC)
+   add_test(NAME confirm_gpu_exists COMMAND check_gpu_num.x)
+else()
+   add_test(NAME confirm_gpu_does_not_exist COMMAND check_gpu_num.x)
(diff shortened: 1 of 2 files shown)


dareg:
Well, some tests need to be duplicated to be run on GPU and CPU or they need to be added to the two list of tests, but it is not sufficient to only run them on GPU. I am for instance thinking of cpu_to_gpu.F90. It is testing that the data transfer works, we want to make sure it works when field_api is used on GPU but also when field_api is run on CPU only. So the test-case needs to be run in the two cases. And it's not the only one. That will lead to a lot of duplicated test-cases with only two lines of pragmas of difference.

Also I am not sure it really needs to be changed now. Because the CI only run on CPU and we are aware of it. So the CI green tick doesn't means that it works on GPU and we know it.  
If, in the future you are able to set a CI on CPU and GPU, then it seems to me that it will be much simpler to compile and run field_api twice, one in a CPU only version and one in a GPU only version. It will test the code correctly without the need of duplicated test-cases.

awnawab:
Thanks for the feedback @dareg! Yes I fully agree that duplicating tests where we are testing both CPU and GPU functionality with only minimal differences would be overkill. We don't have to be super strict here; if a test tests both CPU and GPU functionality, we can simply add the tests regardless of whether OpenACC is enabled (and in doing so recover the behaviour we have now). The idea here is simply to not unnecessarily run tests where we know that the only thing we are testing is GPU offload functionality. Apologies if I wasn't clear enough. I am happy to move any tests you would like out of the OpenACC check.

dareg:
If you want to add a test to check the presence of a GPU, I think the test should only do that, it doesn't need to have anything about field_api, you could only keep the stuff about acc_get_num_devices (don't forget to also change the filename, test's name and test's comment).

awnawab:
Thanks for the feedback @dareg. I've implemented your suggestion, and the name of the test now changes depending on the status of `HAVE_ACC`, so you can tell simply by looking at the ctest log what the intended functionality of the test is.

dareg (approved):

Issue #7, opened 2023-11-10
CUDA backend for faster host-device transfers


awnawab:
I propose the addition of a CUDA backend, that would allow FIELDs to be allocated in pinned host memory and therefore speedup host-to-device transfers by 2x. This would be an addition to the current OpenACC backend. The CUDA backend would be an opt-in (OFF by default). Crucially, backwards compatibility will be maintained; performance and API will be unaffected in the absence of CUDA.

Another possibility afforded by the CUDA backend is the ability to bypass the OpenACC map, thereby removing the need for shadow-host allocations for `FIELD%DEVPTR`. The shadow host allocations are mandated by the need to register discontiguous wrapped FIELDs in the OpenACC map. `FIELD%DEVPTR` therefore becomes a true device-pointer, rather than a host-pointer associated with a device-pointer. This does have an effect on the API. If the CUDA backend is enabled, then we must replace all `!$acc data present` statements with `!$acc data deviceptr` statements.

dareg:
Seems to be a good proposition, but we'll need to see the implementation obviously.
It's also not a problem if using it require to change all the 'present' statement with 'deviceptr', it's all done with a script anyway.

lukasm91:
I agree, this would be very useful. I actually just yesterday implemented a branch where I improve non-contiguous data. I think this might also happen in IFS, though it certainly depends on the memory layout, but I think there are arrays called GMV/GMVS which are of type (NPROMA,NLEV,NFLD,NBLK), and if  we want to copy (:,:,1:20,:), this is very slow. 
For non-contiguous data types, using cudaMemcpy2D instead of the OpenACC cudaMemcpy(1D) equivalent understandably improves performance of those copies by order of magnitude. Of course, this is tangential to this request because pinned memory helps in both cases, but just want to make sure we don't overlap.

awnawab:
Ah that would be an amazing addition, thanks @lukasm91! Because this might affect another development I have in the pipeline for GMV/GFL arrays (will file issue soon for discussion), is cudaMemcpy2D as fast as if the array were contiguous? Or is there still some small performance penalty from copying discontiguous arrays?

lukasm91:
In general, it is only a small penalty, if at all. This is some measurement I took on A100, with a pingpong copy from a pinned buffer to a device buffer. Just ran it fro a few nx/ny/strides and compare the achieved bandwidth between a contiguous 1D copy, and a strided 2D copy.

```
[...]

PR #11, opened 2023-11-15
CUDA backend

awnawab:
This PR implements the feature proposed in #7 and adds a CUDA backend for faster host to device transfers. Backwards compatibility is maintained other than the need to replace `!$acc present` statements with `!$acc deviceptr`. The CUDA backend is enabled by passing `-DENABLE_CUDA=ON` at build time.


dareg:
Do you have some reproducible benchmarks using field_api that show the performance benefits of using the pinning method?
I would like to try them on our clusters.

dareg:
Oh, sorry, I think there was a misunderstanding.
I am not asking you to add benchmarks to the repository.
I just want to know how much improvements you got on your clusters, and how I can reproduce it on our clusters.
You can add the files directly in this conversation thread.

awnawab:
Ah right, sorry about that. This is a very simple test that copies data back and forth. You can add it to the field_api tests and run it with CUDA turned ON/OFF. I saw the total run-time drop from 6.5 to 3.5 when I ran with CUDA enabled.
[cpu_to_gpu_pingpong.txt](https://github.com/ecmwf-ifs/field_api/files/13441187/cpu_to_gpu_pingpong.txt)

awnawab:
FYI @dareg. I have also prepared a real test-case that shows the effect of pinning by creating a variant of CLOUDSC that uses the FIELD_API CUDA backend: https://github.com/ecmwf-ifs/dwarf-p-cloudsc/tree/naan-update-field_api

Once you create the bundle (instructions in the README) and build with the `--with-gpu` and `--with-cuda` options. You can run two variants of CLOUDSC to see the performance impact of the CUDA backend.

With CUDA backend:
```
naan-> NV_ACC_CUDA_HEAPSIZE=12G bin/dwarf-cloudsc-gpu-scc-field 1 262144 128
 Executing CLOUDSC-GPU, "SCC" variant with FIELD API, PACKED STORAGE  F
     NUMPROC=1, NUMOMP=1, NGPTOTG=262144, NPROMA=128, NGPBLKS=2048
 Reference MFLOP count for 100 columns :  12.48232900
     NUMOMP    NGPTOT  #GP-cols     #BLKS    NPROMA tid# : Time(msec)  MFlops/s     col/s
          1    262144    262144         1       128    0 :        171    191142   1531304 @ rank#0:core#195
          1    262144    262144      2048       128   -1 :       1192     27449    219906 : TOTAL @ rank#0
    1 x   1    262144    262144      2048       128   -1 :       1192     27449    219906 : TOTAL
```

Without CUDA backend:
```
naan-> NV_ACC_CUDA_HEAPSIZE=12G bin/dwarf-cloudsc-gpu-scc 1 262144 128
    Executing CLOUDSC-GPU, "SCC" variant...
     NUMPROC=1, NUMOMP=1, NGPTOTG=262144, NPROMA=128, NGPBLKS=2048
 Reference MFLOP count for 100 columns :  12.48232900
     NUMOMP    NGPTOT  #GP-cols     #BLKS    NPROMA tid# : Time(msec)  MFlops/s     col/s
          1    262144    262144         1       128    0 :        171    190907   1529421 @ rank#0:core#67
          1    262144    262144      2048       128   -1 :       2727     11997     96113 : TOTAL @ rank#0
    1 x   1    262144    262144      2048       128   -1 :       2727     11997     96113 : TOTAL
```

For both runs, the kernel performance is around 190 GFs, but with the CUDA backend, the overall performance (including data offload) jumps from 12 GFs to 27 GFs.

pmarguinaud:
Hello Ahmad,

I was wondering whether you could try the following things:

pmarguinaud:
Hello Ahmad,

I was wondering whether you could try the following things:

- Remove the PINNED attribute and apply the cudaHostRegister/cudaHostUnregister directly on %PTR. It is supposed to have the same effect and we avoid the PINNED attribute in the definition of the types
- 
- Remove the DEVICE attribute in the definition of the type, and use the "use_device" clause before calling cuda functions; Lukas is doing that here : https://github.com/lukasm91/field_api/blob/68726f6f55e72056509114f31f1db2c4c273a6e3/field_RANKSUFF_module.fypp#L395C35-L395C39 ; again this would have the effect of minimizing the impact on the implementation but also on the user code where we could keep the present clause

awnawab:
Thanks a lot for reviewing the code and your suggestions @pmarguinaud! Your suggestions have really improved the implementation, and also made it much easier to switch to a HIP backend in the future for AMD GPUs.

Allocations in page-locked memory are no longer achieved using CUDA-Fortran pinned arrays but via `cudaHostRegister` as you suggested. An interesting feature I discovered whilst doing this is that memory allocated via Fortran has an upper limit for pinning for single allocations: very large arrays allocated via `ALLOCATE(SELF%PTR)` cannot be registered in page-locked memory. `cudaHostAlloc` suffers from the same limitation. If however memory is allocated via `C_MALLOC`, then the only upper limit is the OS limit on page-locked memory allocations. This is also why I don't register wrapped fields in page-locked memory, as these will most probably be Fortran allocations.

I also no longer use CUDA-Fortran device arrays and allocate device memory directly on `SELF%DEVPTR`. Again, I have used the CUDA-C API rather than the CUDA-Fortran API so that we can use `TYPE(C_PTR)` rather than `TYPE(C_DEVPTR)` to store the allocation.

The decision to use the `!$acc deviceptr` clause rather than `!$acc present` clause was actually very intentional and not driven by the CUDA API. Currently in FIELD_API, we allocate `SELF%DEVPTR` on the host and use openacc to map this to a device allocation. We need to do this because the openacc map can only register contiguous arrays, and so we cannot register `SELF%PTR` directly in the openacc map because there is no guarantee `SELF%PTR` is contiguous. However if we make `SELF%DEVPTR` a true device pointer, we don't need to register it in the openacc map and we also don't need to do a redundant host allocation for `SELF%DEVPTR`. We can  therefore roughly half the memory footprint on host if we avoid the openacc map and use `!$acc deviceptr` instead of `!$acc present`.

pmarguinaud:
Hello Ahmad,

I am very happy you succeeded in making the implementation simple. Regarding the large memory footprint on the host, could you provide a small test case to illustrate this ? We observed something similar, and solved it with dev_alloc.c

awnawab:
Hi Phillipe,

Please correct me if I am mistaken, but my understanding of dev_alloc.c is that we use a buddy allocator to allocate `SELF%DEVPTR` on host first, and then we use `!$acc enter data create` to map this host allocation to a device allocation. So currently in FIELD_API, `SELF%DEVPTR` is allocated both on host **and** on device.

This extra allocation of `SELF%DEVPTR` on host seems a bit wasteful to me. With `cudaMalloc`, we can allocate `SELF%DEVPTR` directly on device, without needing to allocate anything extra on the host. In this sense, no matter how efficient dev_alloc.c is, it will always require some unnecessary allocations on the host, whereas `cudaMalloc` does not.

I also confirmed this on a real application. The standalone ecWAM dwarf uses ec_meminfo, which provides a very convenient mechanism to track the total heap size. I ran an `O320` grid (~ equivalent to Tco399) both with and without the FIELD_API CUDA backend, and I set `DEV_ALLOC_SIZE` to 2GB. The total heap usage without the CUDA backend was 24433 MB, dropping to 21368 MB with the CUDA backend. So a reduction of more than 10%.

pmarguinaud:
Hello Ahmad,

I am interested in understanding what is happening here. First some explanations on dev_alloc.c: here we allocate a large chunk of virtual memory on the host, which will never be backed by physical memory. This chunk of memory is just allocated to have a mapping between host virtual memory and device memory. So it will never use any physical memory, unless somebody writes into DEVPTR on the host. We added this mechanism, because each time we allocated/deallocated/reallocated/... memory, some physical memory was leaking into DEVPTR members (we saw the physical memory increasing at each time step). 

The total heap size is not the figure to look at; there is also stack memory, and what really matters anyway is physical memory (RSS). 

Now, I would be happy to remove the dev_alloc.c; I do believe you that what you did solves the issue, but we also need to make sure that something similar exists on other architectures. What do you know about this ?

awnawab:
Thanks a lot Phillipe for the explanation on dev_alloc.c, it makes a lot more sense to me now.

For AMD GPUs, we can use the analogous `hipMalloc`:
https://docs.amd.com/projects/HIP/en/latest/.doxygen/docBin/html/group___memory.html

For Intel GPUs, we can use the OpenMP extension `omp_target_alloc_device`:
https://www.intel.com/content/www/us/en/docs/oneapi/optimization-guide-gpu/2023-1/memory-allocation.html
 
(or even the underlying SYCL `malloc_device`)

My suggestion would be to (for the short-term) keep dev_alloc.c as a fallback for when we are running with the CUDA backend disabled. When we later add a HIP (and/or OpenMP) backend for AMD GPUs, then we could possibly remove dev_alloc.c.

@mlange05 your thoughts on this?

pmarguinaud:
Hello Ahmad,

Some other comments:
- memory pinning should not be a compile time option, but should be enabled at run time (eg an option for the field constructor)
- the deviceptr/present could also be a run time option (again a field constructor option)
- I like the deviceptr, it is very neat, when compared to host mapped pointers, and it avoids the need of the complicated dev_alloc.c; but we may have to use some other software where host mapped pointers are expected (eg spectral transforms), so unfortunately, we have to keep both options for now
- please check that dev_alloc.c solves your memory problem on ecwam (look at the rss value only)

awnawab:
Hi Phillipe,

Thank you for the suggestions. As requested, I have now made pinning of host allocations and host-mapped device pointers run-time options. Pinning is enabled by default if CUDA is enabled. Host-mapped device pointers are the default even if CUDA is enabled. If we want true device pointers the argument `MAP_DEVPTR=.FALSE.` has to be passed to the field constructor.

I have also checked the RSS size for ecWAM with dev_alloc.c and with `cudaMalloc`. In both cases it is essentially identical, so dev_alloc.c indeed does solve the memory problem.

field_RANKSUFF_module.fypp
```diff
@@ -201,16 +239,47 @@ CONTAINS
...
+#:endfor
+
+    CALL C_MALLOC(ARR_SIZE, DATA)
 
-    ALLOCATE(SELF%PTR(${', '.join(map(lambda r: 'SELF%LBOUNDS('+str(r+1)+'):SELF%UBOUNDS('+str(r+1)+')', range(0, ft.rank)))}$))
+#:if defined('CUDA')
+    IF(SELF%PINNED)THEN
+      ISTAT = cudaHostRegister(DATA, ARR_SIZE, cudaHostRegisterMapped)
+      IF(ISTAT /= 0)THEN
+          PRINT *, "${ftn}$_OWNER: FAILED TO REGISTER IN PAGE-LOCKED MEMORY"
```
pmarguinaud, line 276: CALL ABOR1

field_RANKSUFF_module.fypp
```diff
@@ -201,16 +239,47 @@ CONTAINS
...
   END SUBROUTINE ${ftn}$_OWNER_INIT
 
   SUBROUTINE ${ftn}$_ALLOCATE (SELF)
     ! Create FIELD object by explicitly allocating new data
+    USE ISO_C_BINDING, ONLY : C_PTR, C_SIZE_T, C_F_POINTER
     CLASS(${ftn}$_OWNER) :: SELF
+    ${ft.type}$, POINTER :: PTR(${ft.shape}$)
+    TYPE(C_PTR) :: DATA
+    INTEGER(C_SIZE_T) :: ARR_SIZE
+    INTEGER :: SHAPE(${ft.rank}$), ISTAT
```
pmarguinaud, line 254: Try to move this in a separate file (allocation of a Fortran pointer with CUDA)

field_RANKSUFF_module.fypp
```diff
@@ -269,9 +338,21 @@ CONTAINS
 
   SUBROUTINE ${ftn}$_OWNER_FINAL(SELF)
     ! Finalizes field and deallocates owned data
+    USE ISO_C_BINDING, ONLY : C_LOC, C_PTR
     CLASS(${ftn}$_OWNER) :: SELF
+    TYPE(C_PTR) :: DATA
+
+    INTERFACE
+      SUBROUTINE C_FREE (PTR) BIND (C, NAME='c_free')
+        IMPORT :: C_PTR
```
pmarguinaud, line 347: Move this in a separate file (deallocation of a Fortran pointer with CUDA)

pmarguinaud (commented):
Hello Ahmad,

Please try to fix the few issues I mentionned. Please also try to make the CUDA stuff non-optionnal: if we are in the PGI environment, it should be available.

awnawab:
Hi Phillipe,

I've implemented your suggestions, the highlight being that CUDA is enabled by default if CMAKE_CUDA_COMPILER is found and OpenACC is also enabled. The changes also highlighted some bugs in the testing, which I've also fixed.

awnawab:
Hi @pmarguinaud,

Happy new year and I hope you enjoyed your holidays 😄  Are there any more changes you would like me to make to this PR? If not, could you please approve it?

pmarguinaud (approved):
This PR enables : 
- the use of device pointers in kernels (with the deviceptr data clause), while keeping compatibility with the existing code (using host mapped pointers). 
- a pinned memory option (on PGI accelerators only for now).

Furthermore, it paves the way for adding support for non NVIDIA GPUs.

PR #16, opened 2023-12-05
Field gang

pmarguinaud:
This branch implements two new classes designed to handle multi-parameter fields such as aerosols. Its purpose is to provide alternative data representation for such fields in the dynamics and in the physics. This is best described in the following presentation:

[fieldbuffer.pdf](https://github.com/ecmwf-ifs/field_api/files/13557021/fieldbuffer.pdf)

The two new classes handle :
- existing fields (eg GMV/GFL) : eg FIELD_3RB_GANG_WRAPPER
- standalone fields (typically temporaries) : eg FIELD_3RB_GANG_OWNER

The implementation relies on a parent/children relationship between a master field of dimension N and children fields of dimensions N-1. All children fields delegate the management of DEVPTR to their parent; the STATUS of the children  is forced to be coherent with the status of the parent.

Other minor cleaning/features are included:
- split some (big) files into smaller ones
- move constants into a separate module
- allow the user to replace ABOR1 by another routine + wrap ABOR1 in FIELD_ABORT


CMakeLists.txt
```diff
@@ -90,27 +90,27 @@ if( NOT fiat_FOUND )
...
-list(APPEND srcs field_basic_module.F90 field_init_debug_value_module.F90 dev_alloc.c)
+list(APPEND srcs field_basic_module.F90 field_init_debug_value_module.F90 dev_alloc.c field_constants_module.F90 field_abort_module.F90)
 
 ## preprocess fypp files
 foreach (SUFF IN ITEMS IM RM RB RD LM)
   string (TOLOWER ${SUFF} suff)
   foreach (RANK RANGE 2 5)
-    foreach (FUNC IN ITEMS "" _gathscat _access _util _array_util)
-      add_custom_command (OUTPUT field_${RANK}${suff}${FUNC}_module.F90
-	      COMMAND ${FYPP} -n -DRANK=${RANK} -DSUFF='${SUFF}' -m os -M ${CMAKE_CURRENT_SOURCE_DIR} -m fieldType
```
awnawab, line 101: Were the fypp line annotations removed intentionally? (the `-n` arg)

field_RANKSUFF_module.fypp
```diff
@@ -237,26 +237,31 @@ CONTAINS
       IF (ZERO) VIEW_PTR(${ft.viewShape}$) = ${ft.default}$
     END IF
 
+    IF (ASSOCIATED (SELF%PARENT)) THEN
+      CALL SELF%PARENT%SET_DEVICE_DIRTY ()
+    ELSE
 ! Mark data dirty on the device: the pointer returned by GET_VIEW may be rw
-    IF (IAND (SELF%ISTATUS, NDEVFRESH)==NDEVFRESH) THEN
-      SELF%ISTATUS = IAND(SELF%ISTATUS, NOT(NDEVFRESH))
+      CALL SELF%SET_DEVICE_DIRTY ()
```
awnawab, line 244: I could be mistaken, but I think here we do not mark the children as dirty on the device if `GET_VIEW` is called on the gang. I think the children should also be marked as dirty.

tests/test_gang.F90
```diff
@@ -0,0 +1,798 @@
...
+      CALL SET_DEVICE_YLF3 ([(JFLD, JFLD = 1, NDIM)])
+    CASE (6)
+      CALL SET_HOST_YLF3 ([(JFLD, JFLD = 1, NDIM)])
+    CASE DEFAULT
+      CALL FIELD_ABORT ('UNEXPECTED CASE')
+  END SELECT
+  
+ELSE
+
+  SELECT CASE (IRAND (1, 134))
```
mlange05, line 144: I'm not sure I understand the use of `IRAND` here. Are we selecting the case to test based on an RNG, or am I missing something here?

CMakeLists.txt
pmarguinaud: Yes and no. Much easier to look at real Fortran code when debugging, but we can got back to fypp line indexing if you do not like it.

tests/test_gang.F90
pmarguinaud: Yes, we generate random (but reproducible) sequences of operations on the 4D fields and its 3D children. Note it is not feasible to explore all the combinations.

field_RANKSUFF_module.fypp
pmarguinaud: The code has moved, but I think SET_DEVICE_DIRTY will call SET_STATUS, which will also update children status.

host_alloc_module.fypp
```diff
@@ -60,39 +62,56 @@ SUBROUTINE ${ft.name}$_HOST_ALLOC (HST, LBOUNDS, UBOUNDS, PINNED)
...
    ${ft.type}$, POINTER, INTENT(INOUT) :: HST(${ft.shape}$)
-   TYPE(C_PTR) :: DATA
+   LOGICAL, INTENT (IN) :: PINNED
 
+   TYPE(C_PTR) :: DATA
+   INTEGER :: ISTAT
+  
    DATA = C_LOC (HST (${ ', '.join (map (lambda i: 'LBOUND (HST, ' + str (i) + ')', range (1, ft.rank+1))) }$))
+
+   IF (PINNED) THEN
```
awnawab, line 102:
```suggestion
#:if defined('CUDA')
   IF (PINNED) THEN
```

host_alloc_module.fypp
```diff
@@ -60,39 +62,56 @@ SUBROUTINE ${ft.name}$_HOST_ALLOC (HST, LBOUNDS, UBOUNDS, PINNED)
...
+     ISTAT = CUDAHOSTUNREGISTER (DATA)
+     IF (ISTAT /= 0) THEN
+       CALL FIELD_ABORT ("${ft.name}$_OWNER: FAILED TO UNREGISTER PAGE-LOCKED MEMORY")
+     ENDIF
+   ENDIF
```
awnawab, line 107:
```suggestion
   ENDIF
 #:endif
```

awnawab (changes requested):
Thanks a lot @pmarguinaud for your implementation of the FieldGang, along with all the other useful additions! Thanks also for adding the missing `CUDAHOSTUNREGISTER`. Once the suggested ifdef guards for that have been added, I am happy to merge this contribution.

tests/test_gang.F90
mlange05: For the record, I don't think RNG-based testing is a good idea, and I still don't fully understand the need for this here. However, fixing this probably requires a rethink of the test base implementation, which is beyond this PR.

mlange05 (approved):
Form a high level this seems ok, but the size of the PR makes a meaningful detailed review very hard. Nevertheless, a good feature to have, so tentatively GTG from me.

awnawab (approved):

PR #19, opened 2024-01-16
Handle 1D arrays with Field API + add options to fieldType.py generator

pmarguinaud:
The purpose of this branch is to add Field API classes for 1D arrays. These classes are similar to exising classes, except that :
- 1D Field classes do not have the GET_VIEW method
- no 1D array class exist
- 1D Field classes have no corresponding gang class
- 1D Field classes cannot be gathered/scattered with the existing gather/scatter object

Some extra attributes have been added to the fieldType object:
- ganged : can this class be part of a gang
- hasView : does this class have the GET_VIEW method

The fieldType generator has been added three extra arguments:
- ganged (True/False), if set keep only classes with matching attribute
- hasView (True/False), if set keep only classes with matching attribute
- alias (True/False), if set keep only classes with matching attribute


fieldType.py
```diff
@@ -28,13 +28,33 @@ def __init__ (self, **kwargs):
...
+    self.hasView = self.rank > 1
+    self.ganged = self.rank > 2
 
 
 kinds = ['JPRM', 'JPRB', 'JPRD', 'JPIM', 'JPLM']
 
-
-def getFieldTypeList (ranks=[2,3,4,5], kinds=kinds):
-  return [fieldType (kind=kind, rank=rank) for (kind) in kinds for rank in ranks]
+def eqv (a, b):
```
awnawab, line 37: [optional style change]

Perhaps this could be simplified a little:
```
def getFieldTypeList (ranks=[1,2,3,4,5], kinds=kinds, hasView=False, alias=True, ganged=False):

  l = [fieldType (kind=kind, rank=rank) for (kind) in kinds for rank in ranks]

  if hasView:
    l = [ft for ft in l if ft.hasView]

  if not alias:
    l = [ft for ft in l if not ft.alias]

  if ganged:
    l = [ft for ft in l if ft.ganged]
```

awnawab (approved):
Thanks @pmarguinaud for these additions! It all looks good to me 👍  The only point I raised is an optional change to `getFieldTypeList`, which I think makes it a little simpler and more readable.

mlange05 (approved):
I agree with the Python style suggestions, but otherwise looks good to me.

fieldType.py
lukasm91: isn't that just

```
l = [ft for ft in l if hasView is not None and hasView == ft.hasView]
```
or
```
if hasView is not None:
    l = [ft for ft in l if hasView == ft.hasView]
```

pmarguinaud: Thank you Lukas, it works !

In Fortran, it is not possible to compare logicals with ==, but with python it works.

PR #25, opened 2024-02-06
0-size allocations for owned fields

awnawab:
Throughout the IFS, we initialize 0-sized owned fields (a workaround for when a particular field is meant to be disabled). Whilst these 0-sized allocations are safe using the Fortran `ALLOCATE` statement, they cause errors in `C_MALLOC`. This PR adds safety checks so that 0-sized allocs can be handled safely using Fortran `ALLOCATE/DEALLOCATE`.

We need `C_MALLOC` because there is an upper limit on registering Fortran allocated memory in page-locked memory (which is lower than the OS limit).


pmarguinaud:
Hello Ahmad,

Why do not you do something like this in c_malloc ?

```
void c_malloc (size_t siz, void ** ptr)
{
  if (siz == 0)
    siz = 1;
  *ptr = malloc (siz);
}
```

This would avoid all the logic with SAFE_ALLOC.

pmarguinaud (approved):

PR #28, opened 2024-02-16
Disable empty generic interfaces

pmarguinaud:
Some generic interfaces (LOAD, SAVE, etc.) are empty : 

```
INTERFACE LOAD
END INTERFACE
```

This is a problem with PGI, which has been reported.  I suggest we just do not generate these empty generic interfaces.


dareg (approved):

mlange05 (approved):
Looks safe to me. GTG

PR #29, opened 2024-02-23
Run with bound checking

pmarguinaud:
Because of this small piece of code (generated from field_RANKSUFF_data_module.fypp) : 

```
  ITHIS_STRIDE = LOC (PTR (LB(1)+1, LB(2), LB(3), LB(4))) - LOC (PTR (LB(1), LB(2), LB(3), LB(4)))
  IF (AFTER < 1) THEN 
    ISIZE = 1
    IF (SIZE(PTR, 1) /= 1 .AND. IPREVIOUS_STRIDE * ISIZE /= ITHIS_STRIDE) THEN 
```

Field API could not run with bound checking enabled : the access to PTR in the LOC intrinsic could go out of bounds when SIZE (PTR, 1) == 1.

This PR solves the issue.

### DIFF
--- a/field_RANKSUFF_data_module.fypp
+++ b/field_RANKSUFF_data_module.fypp
@@ -284,19 +284,22 @@ CONTAINS
   #:endfor
   #:set ind0 = ind0[:-2]
   #:set ind1 = ind1[:-2]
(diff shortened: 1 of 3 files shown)


awnawab (approved):
Well spotted and cleanly fixed! Thanks for this, G2G 👍

PR #34, opened 2024-04-01
Field statistics

pmarguinaud:
The purpose of this PR is to get a few numbers from Field API : 
- current/max memory size managed by Field API
- current/max number of fields created by Field API


tests/test_statistics.F90
```diff
@@ -0,0 +1,70 @@
...
+
+CALL ASSERT (FIELD_STATISTICS%NFIELDS_MAX==2)
+
+CALL ASSERT (FIELD_STATISTICS%MEMORY_DEVICE_CUR==0)
+
+CALL ASSERT (FIELD_STATISTICS%MEMORY_DEVICE_MAX==PRODUCT(INT(UBOUNDS,JPIB))*INT(JPIM,JPIB)+SIZE(IDATA0,KIND=JPIB)*INT(JPIM,JPIB))
+
+CONTAINS 
+
+SUBROUTINE ASSERT (LDCOND)
```
mlange05, line 62: [Optional] This makes test much nicer to read. Could we maybe move this to FIELD_ABORT_MODULE or similar to make this widely available in the test base? Just a thought...

mlange05 (approved):
Looks good to me.

PR #35, opened 2024-04-01
Enable fields with a last lower bound different from 1

pmarguinaud:
The purpose of this PR is to enable fields with a last dimension (usually the block dimension) different from 1; in particular, we test gather/scatter and reshuffle functionnalities on these fields.

Tested in ARPEGE t798, double precision : we run this test case on 12 NVIDIA V100. Using temporaries with block range allowed us to decrease the amount of required memory by a factor of 2. Each MPI task has 72058 = 2252 x 32 points, we process the 2252 blocks in 4 batches of 640 blocks.


awnawab (approved):
Thanks a lot for this @pmarguinaud! This makes the usage of owned fields more flexible and makes the behaviour of the persistent option more coherent 👍

PR #36, opened 2024-05-08
Various cmake fixes

awnawab:
This PR contributes three small CMake fixes:

1. If enabled, cuda is now linked publicly, just like OpenACC. This has the advantage that an externally built/installed FIELD_API with openacc and/or cuda can be linked to shared library targets even if they do not do any offload.
2. A cmake import configuration is added, which exports vital information about the FIELD_API build configuration to any cmake project that issues `find_project( field_api ... )`.
3. FIELD_API can now also optionally use the fypp installation included in fckit's virtual python environment.


dareg (approved):
Tested on our cluster, haven't seen any problem.

PR #37, opened 2024-05-13
Make FIELD_2RB an alias of FIELD_2RD or FIELD_2RM

pmarguinaud:
Using name associations, this PR makes : 
- FIELD_2RB an alias of FIELD_2RD or FIELD_2RM, depending on the precision (single ou double)
- FIELD_2RB_ARRAY an alias of FIELD_2RD_ARRAY or FIELD_2RM_ARRAY

This should make it possible for having a single library for both single/double precision (it we want to). Only a few front-end modules (field_module.fypp, field_array_module.fypp, etc.) would need to be compiled specifically for double/single precision.

This PR prepares the following PR : https://github.com/pmarguinaud/field_api/tree/access-optional

Tested with ARPEGE.


awnawab (approved):
Hi @pmarguinaud, many thanks for this! I prefer the more explicit aliasing of `RB` fields and the cleanup of module imports is also greatly appreciated 🙏 G2G!

PR #38, opened 2024-05-17
Resize only when needed

pmarguinaud:
The resize utility now checks that requested LBOUNDS/UBOUNDS are different from current LBOUNDS/UBOUNDS before resizing.


awnawab (approved):
Very neatly done, G2G!

PR #39, opened 2024-05-17
Add field accessors with optional field argument

pmarguinaud:
Access to field data is made using Fortran generic interfaces such as GET_HOST|DEVICE_DATA_RDONLY|RDWR. Unfortunately, because these generic interfaces are based on functions, they cannot accept an optional field argument.

This PR adds SUBROUTINEs which can be used instead of these functions. For instance : 

```
P => GET_HOST_DATA_RDONLY (YLF)
```

can be replaced by : 

```
CALL SGET_HOST_DATA_RDONLY (P, YLF)
```

The field argument of SGET_HOST|DEVICE_DATA_RDONLY|RDWR routines is optional, which makes it possible to handle such patterns : 

```
SUBROUTINE APL_ARPEGE_PARALLEL (..., YDF)

CLASS (FIELD_3RB), POINTER, OPTIONAL :: YDF
REAL (KIND=JPRB), POINTER :: P (:,:,:)

CALL CALL SGET_HOST_DATA_RDONLY (P, YDF)

```

If YDF is present in APL_ARPEGE_PARALLEL, then P will point to the data of YDF; otherwise, it will point to a dummy array.


awnawab (approved):
Another very nice addition! This has the added benefit that the returned pointer now will have the correct lower bounds 👌 

A side effect of the accessor methods is that the `acc declare create` forces static builds of FIELD API when openacc is enabled. That's not really a problem per se, but it would be nice to have the option of shared builds even if openacc is enabled.

Once the porting activity is done and we are in the optimisation stage, it would be great to evaluate whether we can easily recreate the necessary functionality in user code.

PR #41, opened 2024-05-24
FIELD_ARRAY: safely handle 0-sized fields

awnawab:
This PR adds a check so that FIELD_ARRAY does not wrap 0-sized fields. It also adds the `P_FIELD` pointer member which we need for our Loki recipes to move from thread-local to shared data structures. Finally, there is a small bug fix to the fypp detection mechanism.


pmarguinaud:
Good you thought about zero-sized arrays.

PR #43, opened 2024-05-27
Add proper constructor for array owner

pmarguinaud:
Array owner constructor is not available. This PR tries to fix this.

I have not been able to set up a proper GENERIC INIT interface for both owner/wrapper. Instead, I rely on the presence of the P argument. It would be good,  if someone could look into this.


awnawab:
Hi @pmarguinaud. I filed a PR against your branch to add a generic constructor: https://github.com/pmarguinaud/field_api/pull/2. Either solution is fine with me, you can decide which one to merge.

awnawab:
That confirms that it is indeed a bug. I have tested your fix on GNU 13 and it also solves the problem there 👍

awnawab (approved):

pmarguinaud:
Thank you Ahmad

PR #44, opened 2024-05-29
Reduce compilation time

dareg:
Precision independent files are only compiled once and the few that needs to be compiled twice (the one containing JPRB types) are compiled in single and double precision.
The compilation is then about twice faster.

Please be aware that it will change the name of the library from libfield_api_dp.a to libfield_api.a You might then need to update your build process.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -150,18 +150,47 @@ if( HAVE_ACC )
   ecbuild_warn("OpenACC builds force static linking.")
 endif()
 
+# field api common lib (precision independent)
+ecbuild_add_library(
+  TARGET ${LIBNAME}
+  SOURCES ${srcs}
+  DEFINITIONS $<$<NOT:${fiat_FOUND}>:${FIELD_API_DEFINITIONS}>
+  PUBLIC_LIBS
+    $<${HAVE_ACC}:OpenACC::OpenACC_Fortran>
+  PRIVATE_LIBS
+    $<${fiat_FOUND}:fiat>
+    $<IF:${fiat_HAVE_SINGLE_PRECISION},parkind_sp, parkind_dp>
+    OpenMP::OpenMP_Fortran
+  )
+target_include_directories( ${LIBNAME} PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
+set_property(TARGET ${LIBNAME} PROPERTY C_STANDARD 99)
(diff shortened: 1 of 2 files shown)


awnawab (changes requested):
Hi @dareg. Many thanks for looking into this. Compilation times can indeed be annoyingly long at times.

As you rightly point out, the proposed solution to reduce compile times breaks backwards compatibility with every code that uses FIELD_API. I think we should only break backwards compatibility when absolutely necessary, and we can certainly achieve reduced compile times without doing so.

Furthermore, the proposed implementation also makes it harder to switch between the single and double precision library targets. Where previously we only needed to change the precision suffix, now we must also specify extra includes.

I propose an alternative implementation that addresses both the above: https://github.com/ecmwf-ifs/field_api/tree/naan-dareg-faster-compile

Most sourcefiles are compiled once in `libfield_api`. This is then linked to `libfield_api_dp/sp` which only compiles the few sourcefiles that are precision dependent. I also cleaned up some of the module imports so that the precision dependent sourcefiles don't have to be compiled at all in `libfield_api`.

Please let me know what you think and if this meets your requirements. I have tested this against ifs, ecwam and cloudsc and it works for all three without having to modify the downstream cmake.

pmarguinaud (approved):
Looks good to me. 

Maybe add a libfield_api.a, so that we can (someday) get rid of libfield_api_dp/sp ?

awnawab:
Thanks for your feedback @pmarguinaud. I assume you are referring here to my implementation, since the original proposal already builds libfield_api.a.

It would be fairly easy to enable libfield_api.a to be built, and not building it was an intentional choice. I think as long as FIELD_API defines `FIELD_XRB` (whose definition is dependent upon precision), we should keep the API explicit and keep the precision suffix. The precision suffix sends a clear message to the user that the contents of libfield_api_dp/sp are indeed dependent upon the precision.

If we can instead move the aliasing of `FIELD_XRB` to user-code, then FIELD_API itself will be precision independent and we should then move to a single libfield_api.a.

If you both agree, then @dareg could you please cherry-pick my commit onto your branch and update the PR.

dareg:
Hi @awnawab  ,
I've talked with Philippe and we are okay to accept your branch [naan-dareg-faster-compile](https://github.com/ecmwf-ifs/field_api/tree/naan-dareg-faster-compile), so I will close my PR.
The more important thing for us is to try to reduce the compilation time.

PR #47, opened 2024-05-31
Control data synchronization of field wrappers when device data is created

pmarguinaud:
The following code : 

```
REAL(KIND=JPRB) :: ZZ (NPROMA,NFLEVG,NGPBLKS)
REAL(KIND=JPRB) :: ZP (:,:,:)
CLASS (FIELD_3RB), POINTER :: YLF

CALL FIELD_NEW (YLF, DATA=ZZ)

! Transfer H -> D
ZP => GET_DEVICE_DATA_RDONLY (YLF)
```

would trigger a transfer, even if ZZ is not initialized.

This PR fixes this issue, introducing the INITIALIZED argument of the FIELD_NEW constructor interface : 

```
REAL(KIND=JPRB) :: ZZ (NPROMA,NFLEVG,NGPBLKS)
REAL(KIND=JPRB) :: ZP (:,:,:)
CLASS (FIELD_3RB), POINTER :: YLF

CALL FIELD_NEW (YLF, DATA=ZZ, INITIALIZED=.FALSE.)

! No transfer H -> D
ZP => GET_DEVICE_DATA_RDONLY (YLF)
```


field_RANKSUFF_module.fypp
```diff
@@ -485,6 +502,16 @@ CONTAINS
 
   END SUBROUTINE ${ftn}$_GET_HOST_DATA_RDWR
 
+  SUBROUTINE ${ftn}$_WRAPPER_GET_HOST_DATA_RDWR (SELF, PPTR, QUEUE)
+    CLASS(${ftn}$_WRAPPER) :: SELF
+    ${ft.type}$, POINTER,         INTENT(INOUT) :: PPTR(${ft.shape}$)
+    INTEGER(KIND=JPIM), OPTIONAL, INTENT(IN)    :: QUEUE
+
+    CALL SELF%GET_HOST_DATA (IOR (NRD, NWR), PPTR, QUEUE)
+    SELF%INITIALIZED = .TRUE.
```
awnawab, line 511: `OWNER` fields already have a mechanism to prevent HtoD copies by setting the status to `UNINITIALIZED`. Perhaps I've missed something, but could we not extend that same mechanism to `WRAPPER` fields? This would be cleaner imo rather than having two different mechanisms of checking whether a field is unitialized. This would also allow `GET_DEVICE_DATA` to remain unchanged, rather than duplicating the same code in `WRAPPER_GET_DEVICE_DATA` and `OWNER_GET_DEVICE_DATA`.

pmarguinaud:
Thank you very much for pointing this out. I have followed your advice; the changes are now minimal, and the code looks much simpler.

awnawab (approved):
Thanks a lot for addressing the comment. This looks great now, thanks again for adding this functionality. G2G 👌

PR #49, opened 2024-05-31
Pull back data to legacy structures on device

pmarguinaud:
ARPEGE/IFS field data are now encapsulated with Field API structures in the grid-point calculations; however, in the spectral calculations, these data are still represented by old GMV and GFL arrays.

This implies that we need to implement a fast mean of moving GFL/GMV back and forth between new Field API data structures and GMV/GFL arrays allocated on the device.

This PR implements such a mechanism, which is designed to be flexible enough to be compatible with a traversal of large data structures such as YDVARS.

The test case test_legacy.F90 illustrates this mechanism.


tests/test_legacy.F90
```diff
@@ -0,0 +1,124 @@
...
+U => GET_DEVICE_DATA_RDWR (FU)
+V => GET_DEVICE_DATA_RDONLY (FV)
+T => GET_DEVICE_DATA_RDONLY (FT)
+
+!$acc kernels present (U)
+U = 1._JPRB
+!$acc end kernels
+
+!$acc enter data create (GMV)
+!$acc update device (GMV)
```
awnawab, line 56: [no action] Since GMV has been allocated by the time we get here, we can simply do `!$acc enter data copyin`.

awnawab (approved):
Hi @pmarguinaud. Enabling FIELD_API to work with legacy/manually offloaded code is a much needed and extremely welcome feature addition. It is also very well timed, as we are looking to soon integrate offloaded ecTrans with offloaded ecPhys. Many many thanks 🙌

PR #51, opened 2024-06-04
Faster compilation

awnawab:
This PR builds on top of #44 and reduces compilation time. This is achieved by compiling precision independent sources only once, and sources that alias `FIELD_XRB` are the only ones compiled twice for single and double precision.


awnawab:
Hi @dareg. I would like to tag a minor release in the upcoming days and it would be great to include this too. Could you please give this PR another quick look? Many thanks 🙏

dareg (approved):
Hi,
Just tested it, everything seems to run fine on my side.

PR #52, opened 2024-06-05
Fix test_legacy.F90 (single precision)

pmarguinaud:
Small fix for test this case on single-precision builds


awnawab (approved):
Many thanks for this fix! 🙏

PR #54, opened 2024-06-19
Make GPU offload library linking behaviour configurable

awnawab:
When linking a single FIELD_API build to multiple libraries, we may want to only selectively enable GPU offload in the downstream target libraries.

This PR enables optional private linking of OpenACC and cuda to libfield_api_dp/sp to prevent FIELD_API from always exporting GPU offload libraries to downstream targets.

(I am on leave so will probably be slow to respond to requested changes, apologies in advance!)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -104,6 +104,17 @@ ecbuild_add_option( FEATURE CUDA
     DESCRIPTION "CUDA" DEFAULT ON
     CONDITION CMAKE_CUDA_COMPILER AND HAVE_ACC )
 
+## determine GPU offload library linking behaviour
+ecbuild_add_option( FEATURE EXPORT_GPU_LIBRARY
+                    DESCRIPTION "Export GPU offload library to targets linking to FIELD_API"
+                    DEFAULT ON )
+
+if( HAVE_EXPORT_GPU_LIBRARY )
+   set( _gpu_link_type PUBLIC )
+else()
+   set( _gpu_link_type PRIVATE )
+endif()
+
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -189,8 +200,6 @@ foreach(prec ${precisions})
       TARGET ${LIBNAME}_${prec}
       SOURCES ${prec_srcs} $<TARGET_OBJECTS:${LIBNAME}>
       DEFINITIONS $<$<NOT:${fiat_FOUND}>:${FIELD_API_DEFINITIONS}>
-      PUBLIC_LIBS
```
mlange05, line 192: As discussed offline, I think the fundamental problem is that we want to propagate link flags, but not compiler flags. I tried a suggestion by @reuterbal to switch this to
```
INTERFACE_LIBS
    $<${HAVE_ACC}:OpenACC::OpenACC_Fortran>
```
I think this works for our use case, so potentially the additional configuration option is not needed after all?

mlange05 (changes requested):
Overall, I agree we need to ensure that the compiler flags are not propagated through "PUBLIC" linking of OpenACC / CUDA et al. However, I think we might be able to just switch this to INTERFACE_LIBS to avoid compiler flags seeping upstream, while ensuring that linker flags are propagated correctly. Would you mind verifying this when you're back, @awnawab ?

awnawab:
Closing this as it was superseded by #56.

awnawab:
Closing this as it was superseded by #56.

PR #56, opened 2024-07-17
CMake: Export OpenACC symbols via INTERFACE target linking

mlange05:
_Note: This is an alternative implementation to #54 that solves the same problem._

Recap of the problem: Due to FIELD API exporting the OpenACC target via PUBLIC, we cannot select upstream for which component to enable OpenACC offload and for which to disable it. 

In this implementation we downgrade this to exporting OpenACC via INTERFACE so that all header paths are propagated, but OpenACC can still be selected by the user component. This, in turn requires the internal tests to now explicitly link against OpenACC and set the `-cuda` flag, if enabled (thanks to @awnawab for the fixes; I copied them from #54 ). 

@dareg Could you please confirm that this is backward compatible with the MF build system?  

Edit: Tested and confirmed working with H24 regression testing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -189,8 +189,8 @@ foreach(prec ${precisions})
       TARGET ${LIBNAME}_${prec}
       SOURCES ${prec_srcs} $<TARGET_OBJECTS:${LIBNAME}>
       DEFINITIONS $<$<NOT:${fiat_FOUND}>:${FIELD_API_DEFINITIONS}>
-      PUBLIC_LIBS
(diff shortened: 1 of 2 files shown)


awnawab (approved):
Many thanks for this @mlange05, and apologies for taking so long to review it! This looks really good, and is indeed simpler/cleaner than my implementation of making public linking of openacc configurable. From my side this looks good to go 👌

dareg (approved):
Hi, 
I've just tested it and it seems to compile and run fine on our side.

Issue #55, opened 2024-07-11
Build fails when fckit is used for configuration


pmarguinaud:
Hello,

Our build system adds automatically the fckit project to all Fortran libraries; unfortunately, we have an issue with field_api : 

```
+ set -e
+ cmake -Wno-deprecated -Wno-dev -DCMAKE_C_COMPILER=/home/gmap/mrpm/marguina/gpupack/support/wrap/NVHPC2405/mpicc '-DCMAKE_C_FLAGS= -mp -fPIC -Minfo -gopt -Mlarge_arrays -Mlist -traceback -Mnofma -g -O0 -cuda -acc=gpu -O0 -gopt -gpu=cc70,cc80' -DCMAKE_Fortran_COMPILER=/home/gmap/mrpm/marguina/gpupack/support/wrap/NVHPC2405/mpif90 '-DCMAKE_Fortran_FLAGS= -mp -fPIC -Minfo -gopt -Mlarge_arrays -Mlist -traceback -Mnofma -Mbyteswapio -Mbackslash -Mstack_arrays -g -O0 -cuda -acc=gpu -O0 -gopt -gpu=cc70,cc80 -lstdc++' -DCMAKE_BUILD_TYPE=NONE -DENABLE_SINGLE_PRECISION=OFF -DENABLE_DOUBLE_PRECISION=ON -DENABLE_ACC=ON -DCMAKE_CUDA_COMPILER=/home/gmap/mrpm/marguina/gpupack/support/wrap/NVHPC2405/ccu -DCMAKE_INSTALL_PREFIX=/home/gmap/mrpm/marguina/jetpack/install/FieldAPI -Decbuild_ROOT=/home/gmap/mrpm/marguina/jetpack/install/ecSDK -Dfckit_ROOT=/home/gmap/mrpm/marguina/jetpack/install/ecSDK -Dfiat_ROOT=/home/gmap/mrpm/marguina/jetpack/install/Fiat /home/gmap/mrpm/marguina/jetpack/IAL-external/FieldAPI/field_api
-- The Fortran compiler identification is NVHPC 24.5.0
-- Detecting Fortran compiler ABI info
-- Detecting Fortran compiler ABI info - done
-- Check for working Fortran compiler: /home/gmap/mrpm/marguina/gpupack/support/wrap/NVHPC2405/mpif90 - skipped
[...]

PR #57, opened 2024-07-22
Use fckit contributed fypp even in absence of FCKIT_VENV

awnawab:
This PR fixes the problem identified in #55 and enables FIELD_API to use the fckit provided fypp even in the absence of `FCKIT_VENV` (which requires python3 >= 3.8).


pmarguinaud (approved):
It solves my problem. Good !

PR #58, opened 2024-08-21
Host memory pool

awnawab:
Although pinning host allocations reliably offers faster data transfer times, in some applications the potential gains are reduced by the overheads associated with the more expensive allocation and deallocation. In real applications that will repeatedly allocate/deallocate temporaries, the overheads can be so extreme that the overall walltime actually _increases_ due to pinning. As shown by the two attached graphs, this is indeed the case in the offloaded ecPhys (12-hour forecast on a tco399 grid).

This PR provides a potential solution to this problem in the form of a host memory pool. It comprises of a linked-list of memory blocks, each of which can optionally be pinned. The memory pool can thus grow dynamically, and individual blocks can also be freed for reuse. The actual memory deallocation only happens at the end of the entire run, using a teardown utility called `FIELD_HOST_POOL_DELETE`.

At least in the case of ecPhys, the use of the memory pool almost completely removes the overheads associated with pinning, allowing most of the data transfer walltime gains to be realised.
[ifs_h24_total.pdf](https://github.com/user-attachments/files/16687438/ifs_h24_total.pdf)
[ifs_h24_ecphys.pdf](https://github.com/user-attachments/files/16687439/ifs_h24_ecphys.pdf)


pmarguinaud (approved):

mlange05 (approved):
I really like this, and the implementation is very neat. So, many thanks @awnawa.

For future iterations on this, I note that we could improve the automatic deallocation and page-purging, but for the proof-of-concept that this really great already. GTG for me.

mlange05:
Also, I ran this locally and managed to reproduce performance effect, so going in now....

PR #61, opened 2024-10-01
Fix compilation without FIAT

awnawab:
According to the 2003 Fortran standard, all type-bound procedures of a concrete type must not be deferred. This was not the case when building FIELD_API without FIAT as `CRC64` remained a deferred procedure. The only compiler for which we tested compilation without FIAT was NVHPC, which clearly doesn't enforce the standard strongly in this regard. Compiling with GNU 14 highlighted the bug, which this PR fixes.


dareg (approved):

awnawab:
Thanks a lot @dareg for the quick review 🙏  merging now

PR #62, opened 2024-10-07
Fix propagation of ACC link flags

awnawab:
A small bugfix PR to fix the propagation of acc link flags to executables linking against FIELD_API. This will allow an openacc enabled FIELD_API to be used with a target even if that doesn't link against OpenACC.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -203,8 +203,6 @@ foreach(prec ${precisions})
       TARGET ${LIBNAME}_${prec}
       SOURCES ${prec_srcs} $<TARGET_OBJECTS:${LIBNAME}>
       DEFINITIONS $<$<NOT:${fiat_FOUND}>:${FIELD_API_DEFINITIONS}>
-      INTERFACE_LIBS
-         $<${HAVE_ACC}:OpenACC::OpenACC_Fortran>
       PRIVATE_LIBS
          $<${fiat_FOUND}:fiat>
          $<${fiat_FOUND}:parkind_${prec}>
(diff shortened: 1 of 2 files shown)


mlange05 (approved):
Looks good to me.

CMakeLists.txt
```diff
@@ -203,8 +203,6 @@ foreach(prec ${precisions})
       TARGET ${LIBNAME}_${prec}
       SOURCES ${prec_srcs} $<TARGET_OBJECTS:${LIBNAME}>
       DEFINITIONS $<$<NOT:${fiat_FOUND}>:${FIELD_API_DEFINITIONS}>
-      INTERFACE_LIBS
-         $<${HAVE_ACC}:OpenACC::OpenACC_Fortran>
```
reuterbal, line 207: Do we really not need to link to OpenACC in all cases? I suspect we are getting away with that on NVHPC because the relevant flags added below may automatically imply that behaviour, but CCE or GNU may be handling this differently for example?

reuterbal (approved):
I think you can't assign me because I don't have any write permissions on the repository.

The changes look good to me in principal and I have validated that these work correctly in ifs-bundle. A little unsure whether we should really remove OpenACC from the interface lib, but @wdeconinck might be able to advise

CMakeLists.txt
awnawab: `ecbuild_add_library` doesn't even have an INTERFACE_LIBS option in the API. So INTERFACE_LIBS was in fact being interpreted as a compiler definition. We put that in to avoid F-API exporting openacc compiler flags, and so we were accidentally getting the right behaviour. In any case, a true interface library would indeed export compiler flags too, not just linker flags, so explicitly exporting link flags is the right way to do it.

reuterbal: Thanks, I wasn't aware of this. Looks all good then to me, sorry for the noise!

dareg (approved):

PR #63, opened 2024-10-10
Small field_delete improvement

dareg:
When using field_api inside IAL, we write hundreds of lines similar to this one:
```fortran
IF(ASSOCIATED(F_FOO))CALL FIELD_DELETE(F_FOO)
```

It would make the code simpler if the check was done inside field_delete.


awnawab (approved):
Looks great, a much needed convenience utility to make our lives easier 🙏  Many thanks!

awnawab:
I would like to release a new tag today and it would be nice to include this PR in that tag. I'll merge this at 11:30 AM CET, please shout before then if there are any objections.

PR #65, opened 2024-11-28
Add build-hpc workflow for Github Actions

reuterbal:
This builds with NVHPC on our GPU nodes in single and double precision, each with and without cuda.

### DIFF
--- a/.github/workflows/build-hpc.yml
+++ b/.github/workflows/build-hpc.yml
@@ -1,0 +1,108 @@
+name: build-hpc
+
+# Controls when the action will run
+on:
+
+  # Trigger the workflow on all pull requests
+  pull_request: ~
+
+  # Allow workflow to be dispatched on demand
+  workflow_dispatch: ~
+
+jobs:
+  ci-hpc:
+    name: ci-hpc
+
+    strategy:
+      fail-fast: false    # false: try to complete all jobs
+
+      matrix:
+        prec: [dp, sp]
+        cuda: [False, True]
+        name:
+          - ac-gpu nvhpc
+
+        include:
+          - name: ac-gpu nvhpc
+            site: ac-batch
+            sbatch_options: |
+              #SBATCH --time=00:20:00
+              #SBATCH --nodes=1
+              #SBATCH --ntasks=2
+              #SBATCH --cpus-per-task=32
+              #SBATCH --gpus-per-task=1
+              #SBATCH --mem=200G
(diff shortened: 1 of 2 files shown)


awnawab (approved):
Many many thanks for this 🙏 Running the CI on GPU will be a big boost to the FIELD_API development process. G2G from me 👌

.github/workflows/build-hpc.yml
```diff
@@ -0,0 +1,108 @@
...
+
+jobs:
+  ci-hpc:
+    name: ci-hpc
+
+    strategy:
+      fail-fast: false    # false: try to complete all jobs
+
+      matrix:
+        prec: [dp, sp]
```
awnawab, line 20: By default FIELD_API will build both spdp libraries. Only a few files that are precision dependent are compiled twice, the rest are compiled only once. And all the tests are in fact run in DP, expect for one mixed precision test. So we don't need to extend the matrix to cover DP/SP builds. Sorry for missing this earlier.

awnawab (changes requested):
One small nitpick about the testing matrix that I missed earlier, sorry!

dareg:
Could you give to the workflow a name more similar to the other workflows and more explicit?
Something like FieldAPI_NVHPC_ECMWF_HPC.
Thanks for the PR.

reuterbal:
In my opinion, the other workflows should be condensed into a single workflow with a configuration matrix - that is more in line with what we use in every other repository.

awnawab:
> In my opinion, the other workflows should be condensed into a single workflow with a configuration matrix - that is more in line with what we use in every other repository.

PR #66 filed to do just that 😄

awnawab (approved):
Many thanks @reuterbal for addressing the comment, looks ready to merge to me now 👌

.github/workflows/build-hpc.yml
```diff
@@ -0,0 +1,104 @@
+name: build-hpc
+
+# Controls when the action will run
+on:
+
+  # Trigger the workflow on all pull requests
```
mlange05, line 6: Don't we want to run this on `main` after merge-pushes like we usually do?

reuterbal: Thanks, good point! Fixed now

mlange05 (approved):
Looks great! GTG from me. :shipit:

dareg (approved):

PR #66, opened 2024-11-29
Refactor CI using ecmwf-actions 

awnawab:
This PR brings the FIELD_API CI in line with other software packages by using ecmwf-actions. Coverage is also extended to include:
- nvhpc
- older gnu versions
- macos


.github/workflows/build.yml
```diff
@@ -0,0 +1,214 @@
...
+          - linux gnu-14
+          - linux nvhpc-23.5
+          - linux intel-classic
+          - linux intel-modern
+          - macos
+
+        include:
+
+          - name: linux gnu-10
+            os: ubuntu-20.04
```
reuterbal, line 45: Any particular reason for using the relatively old runner image here? 20.04 is EOL in a couple of months and I expect GH might make these runners unavailable soon after.

.github/workflows/build.yml
```diff
@@ -0,0 +1,214 @@
...
+            os: ubuntu-24.04
+            compiler: gnu-14
+            compiler_cc: gcc-14
+            compiler_cxx: g++-14
+            compiler_fc: gfortran-14
+            python-version: '3.11'
+            caching: true
+
+          - name: linux nvhpc-23.5
+            os: ubuntu-20.04
```
reuterbal, line 63: Same (also for intel below)

.github/workflows/build.yml
```diff
@@ -0,0 +1,214 @@
+name: build
+
+# Controls when the action will run
+on:
+
+  # Trigger the workflow on all pushes, except on tag creation
+  push:
+    branches:
+    - '**'
```
mlange05, line 9: Can we restrict the runs on push to the `main` branch only, please? Pull requests are already enabled below, as well as manual workflow dispatches for early testing.

.github/workflows/build.yml
awnawab: Great catch!

mlange05 (approved):
Looks great to me, very neat! GTG from me. :shipit:

.github/workflows/build.yml
dareg: I disagree with that. Having the tests run by themselves for any commit when experimenting on a branch is very useful and should not be deactivated. It's better to be aware of a problem as soon as possible than at the final step when opening a pull request.
And having to run the tests by hand seems to defeat the purpose of continuous integration.

dareg: Yes I would like to keep it for the public GH-hosted runners please.
And I understand you don't want to overload your GPU-queue.

dareg (approved):

PR #67, opened 2024-11-29
Add CPU only support for CCE17

awnawab:
This PR provides several fixes needed to build and use FIELD_API using the Cray Fortran compiler. Note that GPU offload is still only supported for NVHPC; Cray OpenACC support will be added at a later date.

It is a little surprising how many problems Cray Fortran had with FIELD_API given that it works with gnu, intel (old and llvm) and nvhpc. Nevertheless, the fixes are reasonably clean. The one exception is the FieldGang, for which the cray compiler cannot resolve the polymorphism, so I simply disabled it. Interestingly, cce15 can build the FieldGang but not newer versions. cce15 was beset with many other problems though, so the extra complexity of enabling the FieldGang for cce 15 or older would not be worth it.


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf-ifs/field_api?pullRequest=67) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf-ifs/field_api?pullRequest=67) before we can accept your contribution.<br/>**1** out of **2** committers have signed the CLA.<br/><br/>:white_check_mark: awnawab<br/>:x: MichaelSt98<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf-ifs/field_api?pullRequest=67) it.</sub>

awnawab:
FYI @ioanhadade

field_RANKSUFF_gather_module.fypp
pmarguinaud, line 1: What is the problem with generic INTERFACE here ? Why did you have to move their definition in other files ?

awnawab: We were previously defining a generic interface in each field_RANKSUFF_gather_module, and then importing these interfaces in field_gathscat_module. When building field_gathscat_module, the cray compiler complained that the same symbol (the generic interfaces all had the same name of course) is defined in multiple modules. The cleanest fix I could think was to simply move the definition of the interface into field_gathscat_module.

ioanhadade:
> FYI @ioanhadade

Thanks, can confirm this builds with cce 16.0.1 and 17.0.1

CMakeLists.txt
```diff
@@ -61,7 +61,8 @@ endif()
 ecbuild_add_option( FEATURE ACC
                     DEFAULT ON
                     DESCRIPTION "Support for using GPUs with OpenACC"
-                    REQUIRED_PACKAGES "OpenACC COMPONENTS Fortran" )
+                    REQUIRED_PACKAGES "OpenACC COMPONENTS Fortran" 
+                    CONDITION CMAKE_Fortran_COMPILER_ID MATCHES "PGI|NVHPC")
```
mlange05, line 65: I think we want to make this explicit in the README under supported features.

CMakeLists.txt
```diff
@@ -124,6 +125,13 @@ ecbuild_add_option( FEATURE BUDDY_MALLOC
     DEFAULT ON
 )
 
+## Field Gang
+ecbuild_add_option( FEATURE FIELD_GANG
+    DESCRIPTION "Enable packed storage of fields"
+    DEFAULT ON
+    CONDITION (NOT CMAKE_Fortran_COMPILER_ID MATCHES "Cray")
```
mlange05, line 132: Same here, should we mention this in the "supported features" section of the README?

mlange05 (changes requested):
Looks good to me (untested though). Only comment I have is that we should probably update the README to reflect the changes in outward-facing CMake options.

mlange05 (approved):
Looks great now from my side. 👍

pmarguinaud (approved):

field_RANKSUFF_gather_module.fypp
awnawab: Yes I agree, and in fact we have the same pattern in field_factory_module which the compiler seems to accept. Good idea, I will make a reproducer and share it with HPE.

dareg (approved):
I don't have access to a cray compiler so I've just checked that it still works with our gpupack setup.

awnawab:
Thanks all for the feedback and reviews 🙏  merging now

Issue #68, opened 2024-12-04
Add support for partial offloading of field API fields to device


wertysas:
### Is your feature request related to a problem? Please describe.

When working with large fields they may not fit on device. A well-known solution, when running out of device memory,  is to block the data offload and device kernels in a loop and operate on smaller chunks. However, there is no ability to partially offload Field API fields onto device without allocating the entire field on device. This makes it impossible to block device offloads when using Field API without introducing new fields for the blocks, which leads to unnecessary allocations.

### Describe the solution you'd like

I propose that we introduce partial offloading of fields to device. This will allow blocking of device kernels, without running out of memory on device for large fields. An implementation of such blocking using OpenACC and fortran arrays has been demonstrated to work on [dwarf-p-cloudsc](https://github.com/ecmwf-ifs/dwarf-p-cloudsc) with both increased performance and the possibility to handle much larger problem sizes than before without running out of device memory(see more below). 

_I suggest a series of three pull request to introduce partial offloading of fields to Field API:_

**PR 1:** Make the lower bounds of the ``DEVPTR`` member of fields always start from one, as opposed to before, when it inherited the lower and upper bounds of the ``HST`` pointer member. This change is a first step towards decoupling the bounds of the device pointer from the bounds of the host pointer. **Note that this does not change the user-facing API.**

**PR 2:** Introduce the ability to offload *blocks* of fields onto the device and only allocate the block size on device. With the use case in mind, I think it makes sense to only allow blocking in the outermost dimension. This could be achieved by introducing an optional dummy argument, ``BLK_BOUNDS``, to the copy methods of fields. This would be an array of two integers that hold the lower and upper index of the block. An example of how the updated signature  would look in the ``GET_DEVICE_DATA`` method is shown below:
```
SUBROUTINE ${ftn}$_GET_DEVICE_DATA (SELF, MODE, PTR, QUEUE, BLK_BOUNDS)
```
The following is a toy example of how blocked offloading and kernel invocations could look with the suggested API.
```fortran
NUM_BLKS=10
DO BLK=1,10
  BLK_BOUNDS = [(BLK-1)*NUM_BLKS+1, BLK*NUM_BLKS]
  CALL W%GET_DEVICE_DATA_RDWR(W_DEVPTR, BLK_BOUNDS=BLK_BOUNDS)
  ! kernel parallelised over BLK_BOUNDS(1) to BLK_BOUNDS(2)
[...]

PR #69, opened 2024-12-04
Make ``FIELD%DEVPTR`` LBOUNDS always one

wertysas:
_This PR is the first in a series of PR's related to the issue [Add support for partial offloading of field API fields to device #68](https://github.com/ecmwf-ifs/field_api/issues/68)._

This PR decouples the bounds of ``FIELD%DEVPTR`` from ``FIELD%PTR`` member, by always making the ``FIELDS%DEVPTR`` lower bound be one. The decoupling of the device and host pointers is the first step towards partial field offloads, as outlined in the issue linked above.


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf-ifs/field_api?pullRequest=69) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf-ifs/field_api?pullRequest=69) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf-ifs/field_api?pullRequest=69) it.</sub>

awnawab (approved):
Many many thanks @wertysas for this very neatly implemented first step of #68 👌 This is good to go from my side, but as this is quite a fundamental change, testing this thoroughly is essential. @mlange05 could you please confirm this branch doesn't break ecPhys offload? @dareg could you also please confirm these changes are safe on your end?

mlange05 (approved):
Changes look good to me and I can confirm that EC-physics regression test works fine with this on GPU.

pmarguinaud:
Hello,

I would like this to be an option, as we already rely on %DEVPTR having the same bounds as %PTR in our physics.

awnawab:
Ah, I didn't know we had use cases where external code could access `SELF%DEVPTR` directly. In that case of course we can make this optional. Thanks!

pmarguinaud:
The best solution would be to use a local pointer with lower bounds equal to one; is not that possible ?

awnawab:
Hi Phillipe,

There's two features we are hoping to add:

1. Reduce the memory footprint on device
2. Break up the data-transfers and kernels into several chunks, and run these asynchronously on device so as to overlap communication and computation

If we are to use the `DEVPTR` storage pointer in F-API (which we would like to to minimise complexity), then achieving the above requires `HST` and `DEV` in `${ftn}$_COPY_DIM${d}$_CONTIGUOUS` to have different bounds.

I assumed the storage pointer was not part of the external API (in fact I wondered why we haven't marked it as private) so I advised Johan to go with the simplest solution: assume `DEV` has lbounds of one and map the bounds of `HST` to `DEV` accordingly. Apologies for that Johan, my misunderstanding has created a bit of extra work for you.

If I am understanding your proposal correctly, you are suggesting to place a device storage pointer in user code outside of F-API and use that to do the copy? If so, we could achieve point 1 with that approach, although for fields used in multiple kernels (e.g. `YDVARS`) it would be very hard to do without commiting device storage pointers to the IFS (which I think we must avoid). The second of our aims, which requires all the machinery for asynchronous copies, I would really like to keep the device storage inside F-API.

With the new knowledge of your physics use-case, @wertysas and I will sit together offline and come up with a clean workaround.

awnawab:
Hi Philipe,

We want to remove the limitation that `SELF%DEVPTR` and `SELF%PTR` always have to be of the same size (of course this will be optional, and the default will be the current behaviour). To remove this limitation, the 1D copy methods have to account for potentially different bounds of `PTR` and `DEVPTR` and compute the necessary mapping. This mapping is simplest when `DEVPTR` has lbounds of 1. Of course we can also compute the mapping even if the lbounds of `DEVPTR` are not 1. Johan is now working to remove the hard-coding of the lbounds to 1, and is instead working to update the 1D copy method so that it simply checks the lbounds of `HST` and `DEV` and computes the mapping. The 2D copy method is already written in a way that it does not rely on `HST` and `DEV` having the same bounds, and so it should be straightforward to do the same in the 1D copy method. We'll drop you a line once Johan has applied those changes and we'll mark the PR ready for review again.

awnawab:
Thanks a lot @wertysas for addressing the comments 🙏 @pmarguinaud this is now ready for review again, please have a look whenever you can.

pmarguinaud (approved):

dareg:
Hello,
Could you make sure all the tests pass please

mlange05:
Hi @dareg , yes we'll make sure we get tests green first. The current `ci-hpc` failures are due to some outdated ssh keys on the test runners. We've flagged this with the respective admins but will have to wait for them to update the runners. I'll wait to get the test to pass before merging though.

dareg:
Okay, thanks. Sorry to be nit-picking like that. I'm afraid that if we allow a PR with a failing test today, we might have PR with multiple failing tests in a year.

mlange05:
Hi @dareg, I fully agree. The runners have been fixed now and all tests are passing.

PR #71, opened 2024-12-18
BUGFIX: Add missing field_statistics_module imports in dev_alloc_module

wertysas:
This PR fixes a bug in ``DEV_ALLOC_MODULE``. It adds two missing imports of ``FIELD_STATISTICS_MODULE`` that were making compiles without CUDA or BUDDY_MALLOC enabled to fail.


awnawab (approved):
Great catch and very neatly fixed, G2G from me 👏

dareg (approved):

PR #75, opened 2025-01-28
Move FIELD_GANG into buffer subdir

awnawab:
A further small reorganisation of the code which moves the `FIELD_GANG` into a `buffer` subdir. The constructor/destructor for the `FIELD_GANG` is also moved to the `buffer` subir by splitting `FIELD_RANKSUFF_FACTORY_MODULE` into submodules. Submodules are how I plan on integrating the optional hdf5 I/O functionality so it would be good to test the submodule pattern before that work begins in earnest.


pmarguinaud (approved):

mlange05 (approved):
Looks great, GTG.

dareg (approved):

PR #78, opened 2025-01-28
Run hpc-ci on PRs filed from forks

awnawab:
This PR adds the ability to run the hpc-ci on PRs filed from forks. Importantly, this will only run once a maintainer has added the "approved-for-ci" label. Any further pushes to the PR will remove the label, and it will have to be manually re-added to re-run the hpc-ci. Thanks a lot @samhatfield for showing us how to do this 🙏

### DIFF
--- a/.github/workflows/build-hpc.yml
+++ b/.github/workflows/build-hpc.yml
@@ -16,9 +16,14 @@ on:
   # Allow workflow to be dispatched on demand
   workflow_dispatch: ~
 
+  # Trigger after public PR approved for CI
+  pull_request_target:
+    types: [labeled]
+
 jobs:
   ci-hpc:
     name: ci-hpc
+    if: ${{ !github.event.pull_request.head.repo.fork && github.event.action != 'labeled' || github.event.label.name == 'approved-for-ci' }}
 
     strategy:
       fail-fast: false    # false: try to complete all jobs
--- a/.github/workflows/label-public-pr.yml
+++ b/.github/workflows/label-public-pr.yml
@@ -1,0 +1,10 @@
+# Manage labels of pull requests that originate from forks
+name: label-public-pr
+ 
+on:
+  pull_request_target:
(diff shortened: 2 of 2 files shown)


reuterbal (approved):

reuterbal (changes requested):
Just yesterday we realized that for the `pull_request_target` trigger, this action checks out the wrong branch because `${{ github.sha }}` will point to the target branch top commit instead of the HEAD of the pull request branch.

We haven't confirmed this, yet, but it might be that we need to use `${{ github.ref }}` instead (in l. 106 of your workflow file).

awnawab:
Thanks @reuterbal for picking up on that! I'm turning this back into a draft whilst I investigate the above.

samhatfield:
`github.ref_name` is another contender. I think we just have to put some print statements in to the workflow file to figure it out. There is also `github.event.pull_request` which may have all the properties we need.

awnawab:
To test the above I pushed a commit with a cmake debug print, and this print does indeed show up in the hpc-ci. The commit hash printed in the hpc-ci is however not the same as the HEAD of the PR branch, but that is explained by the following:
```
github.sha
This is the SHA for a temporary commit created for validating the pull request. The commit represents the results of a point-in-time merge of the pull_request.head to the pull_request.base. This value is also exposed through the environment variable GITHUB_SHA. By default, actions/checkout will pull this commit into the local workspace.
```

So it seems like `github.sha` does indeed correspond to the correct code, but the commit id is different because `github.sha` represents the merge commit of the PR head and PR base.

reuterbal (approved):
Thanks for this, I fell into a trap with the different hash there. All good then!

awnawab:
Thanks @reuterbal  and @samhatfield  for the feedback! @dareg could you please have a look at this at your earliest convenience? I would like to merge this and #75 soon if you approve.

dareg (approved):

Issue #73, opened 2025-01-21
Adding support for multiple offload backends


awnawab:
Currently, FIELD_API only supports GPU offload via Nvidia's flavour of OpenACC as well as using the cuda runtime API in certain places. If we are to meaningfully expand the support for other offload backends (e.g. OpenMP or even Cray OpenACC) continuing in the current manner (i.e. ifdefs) is unsustainable.

I propose the following two potential solutions to expand offload backend support:

### 1. Macros

Right now, we directly write openacc pragmas and openacc/cuda runtime functions in the fypp files. Instead, these should be replaced by macros.

For example, the instruction to copy a contiguous chunk of memory to device currently is:
```
CALL ACC_MEMCPY_TO_DEVICE (DEVPTR , HST (${ar}$), ISIZE)
```
It should instead become:
```
$:COPY_TO_DEVICE_1D(DEVPTR , HST (${ar}$), ISIZE)
```

Each offload backend would then have an appropriate implementation for `COPY_TO_DEVICE_1D`. The question then arises as to how the various backend implementations should be defined. The simplest approach that would lead to the least amount of code repetition would be to implement the backends as python modules. Using python modules will enable the use of polymorphism wherever appropriate, e.g., `class NvidiaOpenaccCuda` would be an extension of `class NvidiaOpenacc`.

### 2. Replicating the necessary files

It is only three files that contain GPU offload related instructions:
- field_RANKSUFF_data_module.fypp
- dev_alloc_module.fypp
- host_alloc_module.fypp

We could simply create a copy of each of these files for each backend we are interested in.

Whilst the code may be slightly more readable with solution 2, it will definitely lead to a lot more code replication. Primarily for this reason, I am leaning strongly towards solution 1.

As this would be a big change to FIELD_API, I would really love everyone's input on the above proposal @dareg @pmarguinaud @mlange05 @wertysas.

NB: Both the current issue and issue #72 would benefit greatly from a more logical directory structure of FIELD_API rather than the flat one we currently have. So whilst we discuss the above I will file a PR to that end.

awnawab:
I've since discovered that whilst in the "core" library only three files contain offload related instructions, a lot of the "utilities" also contain offload instructions. So there are a lot more than 3 files that contain offload instructions. This further strengthens the argument for using macros, as suggested in solution 1.

PR #80, opened 2025-01-28
Implement offload instructions  via macros

awnawab:
A draft PR that represents a proof of concept of the development I proposed in #73.

The offload instructions in the code are replaced with macros, which are defined in python modules. More of these python modules can be added to support multiple backends, whilst leaving the code relatively clean. So far I have only implemented the macros for the code in `FIELD_RANKSUFF_DATA_MODULE` that would be run if OpenACC is enabled but not CUDA.

The preprocessor definitions need to still be generalised. I envisage the following two: `WITH_GPU_OFFLOAD` corresponding to `_OPENACC` and `WITH_HIC` corresponding to `_CUDA`.

I am eager to start work (very) soon on an OpenMP offload backend, but before I write more code I would really appreciate your feedback on this.


awnawab:
NB: the hpc ci tests are failing because of a permissions issue for PRs filed from forks. Once #78 is merged I can rebase on top of that to fix the failing hpc-ci. I have tested offline that it works.

cmake/field_api_find_fypp.cmake
```diff
@@ -26,16 +26,14 @@ macro( field_api_find_fypp )
...
-   elseif( FYPP MATCHES FYPP-NOTFOUND )
-     include(cmake/field_api_fetchcontent_fypp.cmake)
-     set(FYPP ${fypp_SOURCE_DIR}/bin/fypp)
-     ecbuild_info("fypp downloaded to: ${FYPP}")
+   elseif( FYPP MATCHES FYPP-NOTFOUND OR (fckit_FOUND AND (NOT fckit_HAVE_FCKIT_VENV OR NOT DEFINED fckit_HAVE_FCKIT_VENV)) )
+     # Discover only system install Python 3
+     set( Python3_FIND_VIRTUALENV STANDARD )
+     find_package( Python3 COMPONENTS Interpreter )
+
+     execute_process( COMMAND ${Python3_EXECUTABLE} -m pip --disable-pip-version-check install fypp OUTPUT_QUIET )
```
reuterbal, line 34: Small comment from something we've hit very recently: Bare-metal Python installations may come without any pip or only ancient versions available. But they are guaranteed to have a bootstrap mechanism for it that ensures that you have at least a pip version equivalent to the ensurepip version packaged with the installation. For this, add the following:

```suggestion
     execute_process( COMMAND ${Python3_EXECUTABLE} -m ensurepip --upgrade OUTPUT_QUIET )
     execute_process( COMMAND ${Python3_EXECUTABLE} -m pip --disable-pip-version-check install fypp OUTPUT_QUIET )
```
```

File: python_utils/offload_backends/nvidia/openacc.py [hunk 2]
```diff
@@ -0,0 +1,91 @@
...
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+
+__all__ = ['NvidiaOpenACC']
+
+class NvidiaOpenACC():
```
reuterbal, line 13: The idea of categorising backends by compiler-specific programming model implementations is a really good one, it allows to easily encode the minor differences in the interpretation of standards between vendors.

I would suggest to name them according to compiler toolchain, though, renaming this one e.g. to `NVHPCOpenACC`. This avoids also any issues associated with putting protected brand names like Nvidia into code.

reuterbal (commented):
This is very neat and I like how clean this looks, despite the powerful features it unlocks for us. While discussing this offline we've noticed two small things that I've marked as comments.

python_utils/offload_backends/nvidia/openacc.py
awnawab: Yes I agree, thanks!

cmake/field_api_find_fypp.cmake
awnawab: I didn't even know we could have pip-less python, thanks!

awnawab:
This is now ready for review, please have a look at your earliest convenience 🙏

pmarguinaud (commented):
I agree we need something like this, but the code which was already difficult to understand, is becoming cryptic. 

Would it be possible to translate OpenACC directives such as : 

!$acc kernels present (PX)

Into : 

$:offload_macros.kernels (present=['PX'])

So that it keeps its meaning in the NVIDIA lingo.  I know it looks unfair for other vendors, but we have to choose a convention anyway.

awnawab:
Hi @pmarguinaud,

Of course, if you think that makes the code more intuitive I am very happy to adapt it accordingly. Thanks for the feedback, I'll implement your suggestion 👍

awnawab:
Hi all. Thanks again for the feedback so far. I would really like to close this soon, so could you please give this another look and let me know what you think? Thanks!

mlange05 (approved):
Looks very good to me and gets us a huge step towards multi-architecture support - many thanks! Ready to go for me.

reuterbal (approved):
I think this looks really good!

wertysas (approved):
I think it looks very good. This will make it much easier to support multiple backends.

dareg (approved):

PR #83, opened 2025-02-12
Add support for blocked offload of fields to device

wertysas:
This PR adds the ability to offload partial fields, referred to as *blocks*, to the device. This corresponds to **PR2** described in the issue [Add support for partial offloading of field API fields to device #68](https://github.com/ecmwf-ifs/field_api/issues/68).

A new optional dummy argument ``BLK_BOUNDS``, an array of two integers that hold the lower and upper index of the block to be offloaded, is added to all data transfer methods. If the field is not already allocated on device the new allocation of the field on the device will only be the size of the *block*.

The implementation has been tested on the [cloudsc-dwarf](https://github.com/ecmwf-ifs/dwarf-p-cloudsc/tree/je-field-offload-blocked) on GPU nodes (NVIDIA A100) of ECMWF's Atos supercomputer by running instances with 10x larger fields than what is possible without blocking. The results from such an experiment is shown in the plot below. The blocked version ``SCC-FIELD`` is not only able to handle larger problem loads but also displays better performance than the non-blocked OpenACC variant ``SCC``.

partial-offload


src/core/field_RANKSUFF_module.fypp
```diff
@@ -573,65 +604,70 @@ CONTAINS
...
-
-    CALL SELF%GET_DEVICE_DATA (IOR (NRD, NWR), PPTR, QUEUE)
+    INTEGER(KIND=JPIM), OPTIONAL, INTENT(IN)    :: BLK_BOUNDS(2)
+   
+    CALL SELF%GET_DEVICE_DATA (IOR (NRD, NWR), PPTR, QUEUE, BLK_BOUNDS=BLK_BOUNDS)
 
   END SUBROUTINE ${ftn}$_GET_DEVICE_DATA_RDWR
 
-  SUBROUTINE ${ftn}$_SYNC_DEVICE_RDWR (SELF, QUEUE)
+  SUBROUTINE ${ftn}$_SYNC_DEVICE_RDWR (SELF, QUEUE, BLK_BOUNDS)
```
awnawab, line 667: We need to pass the `BLK_BOUNDS` through here too.

tests/test_get_device_data_bounds.F90
```diff
@@ -0,0 +1,79 @@
...
+
+  IF (.NOT. OKAY) THEN 
+    CALL FIELD_ABORT("ERROR HOST DATA NOT UPDATED BY SYNC_HOST_RDWR")
+  END  IF
+
+  DO I=1,128
+    IF ( PTR_CPU(I,3) /= 37 ) THEN
+      OKAY =.FALSE.
+    END IF
+  END DO
```
awnawab, line 72: I would add here another offload round but this time for the full field, to test the `CREATE_DEVICE_DATA` mechanics.

awnawab (changes requested):
Thanks a lot @wertysas for this excellent piece of work 🙏 You've added very useful new functionality in a super clean way, well done! Kudos also for taking what must have been a difficult rebase in your stride 😅 

I've suggested a few minor fixes and one more significant one to the `CREATE_DEVICE_DATA` logic. It would also be good to expand the testing to cover the latter.

src/buffer/field_RANKSUFF_gang_module.fypp
```diff
@@ -64,8 +64,9 @@ PUBLIC :: ${ftn}$_GANG_${type}$
 
 CONTAINS
 
-  SUBROUTINE ${ftn1}$_CREATE_DEVICE_DATA_WRAPPER_HELPER (SELF)
+  SUBROUTINE ${ftn1}$_CREATE_DEVICE_DATA_WRAPPER_HELPER (SELF, BLK_BOUNDS)
     CLASS(${ftn1}$_WRAPPER_HELPER) :: SELF
+    INTEGER(KIND=JPIM), OPTIONAL,  INTENT(IN) :: BLK_BOUNDS(2)
```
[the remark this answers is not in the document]
wertysas: Thank you for catching this! I've fixed that now.

src/core/field_RANKSUFF_module.fypp
```diff
@@ -479,68 +499,79 @@ CONTAINS
...
     INTEGER (KIND=JPIM), OPTIONAL, INTENT(IN)    :: QUEUE
+    INTEGER(KIND=JPIM), OPTIONAL,  INTENT(IN)    :: BLK_BOUNDS(2)
+    
     INTEGER(KIND=JPIM)                           :: LBOUNDS(${ft.rank}$)
-
+    
     LBOUNDS=LBOUND(SELF%PTR)
     IF (.NOT. ASSOCIATED (SELF%DEVPTR)) THEN
-      CALL SELF%CREATE_DEVICE_DATA
+      CALL SELF%CREATE_DEVICE_DATA(BLK_BOUNDS=BLK_BOUNDS)
```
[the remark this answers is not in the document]
wertysas: Agreed, I have moved the logic to the ``CREATE_DEVICE_DATA`` routine now and added the missing delete.

tests/test_get_device_data_bounds.F90
wertysas: Yes, that should definitely be covered by the test, I have added another offload round that triggers the "reallocation".

awnawab:
Thanks a lot again for this @wertysas. After a lengthy offline discussion with @mlange05 and @wertysas, we realised that the current implementation has an inconsistency when it comes to the status tracking; a partial offload of a field shouldn't set the entire field dirty/fresh on host/device.

Furthermore, we realised that the existing asynchronous functionality is also incompatible with the status tracking. Consider the following example which would lead to undefined behaviour:

```
CALL FIELD%SYNC_DEVICE_RDWR(QUEUE=1)
CALL FIELD%SYNC_HOST_RDWR()
```

The status tracking should therefore only be reserved for the default use case of offloading entire fields synchronously. Anything more complex than that, e.g. asynchronous or partial offload, should go via an "expert" mode, where the user is able to override the status checks.

As such, @wertysas could you please prepare a PR where you enable this user override for the field status tracking? This PR should be converted to a draft and rebased once the expert mode functionality is merged.

PR #86, opened 2025-02-25
ACC_MEMCPY_TO_DEVICE_ASYNC bugfix

awnawab:
The recent rewrite of the offload instructions using macros introduced a bug to the asynchronous copies. This PR fixes it.


dareg (approved):
Good catch.
I think this bug shows that we are in need of more test-cases. Some that would use the new macro system, in addition to the current set of test-cases.

mlange05 (approved):
Looks good to me, and confirmed working in local regression test. GTG.

awnawab:
Thanks both for the feedback, and fully agreed @dareg on the need for more testing 🙏

PR #87, opened 2025-03-05
Safely transfer 0-sized allocations

awnawab:
A small PR to add checks to enable safe transfers of data between host and device of 0-sized allocations.


mlange05 (approved):
Looks great, many thanks.

PR #89, opened 2025-03-18
Refactor asynchronous offload using CUDA streams

awnawab:
This is the first in a series of PRs aiming to add OpenMP offload support.

OpenACC uses the streaming model to do asynchronous offload, and offers the user fine-grained control over which operations to associate to which streams (queues). OpenMP follows the tasking model but doesn't allow the user explicit control over which operations to associate with a task (or streams/pipelines of tasks). Dependencies between tasks can be expressed via dependencies on variables, but this is less generalisable.

Fortunately, OpenMP offload is interoperable with both CUDA and HIP, both of which have native streaming interfaces. This PR refactors asynchronous offload to make it more portable by explicitly using cuda streams. For OpenACC, each stream is mapped to a queue. For OpenMP, native streams are created. The user API remains the same in either case; the user controls the streams via the `QUEUE` parameter.

The only downside of the current implementation is that asynchronicity now requires the cuda/hip backend, although truly asynchronous copies require the host memory to be pinned anyway, which can't be done without the CUDA/HIP backends.

The next PR after this will add an openmp and openmp+cuda backend for nvhpc (naan-openmp-offload). I also have an openmp and openmp+hip backend for amd's flavour of flang, but I will wait to PR this until some compiler bugs are fixed.


awnawab:
Thanks for the feedback @dareg. In ecWAM we now have the entire timestep on device, but for things that still run on host e.g. I/O, nemo coupling, I split the pullback into multiple queues and only do the wait on the queue that we need. So it would be good to keep the option for multiple queues for OpenMP as well.

https://github.com/ecmwf-ifs/ecwam/blob/bd552cb118a56c18a83c4ee028530bfaa2ee599b/src/ecwam/wamintgr_loki_gpu.F90#L190

pmarguinaud (approved):

mlange05 (approved):
Looks great! Many thanks, GTG. :shipit:

PR #90, opened 2025-03-20
Add clone functionality to a field

dareg:
This PR adds a way to clone a field into a second field, which will always be a field owner.
We are already using this functionality in [IAL](https://github.com/ACCORD-NWP/IAL/blob/7b7f1b20f0618d2c7d17cd5798b4d1601e866c42/arpifs/module/yomradf.fypp#L175), but only on CPU.
We are now proposing to upstream an improved version (for CPU and GPU) of this functionality into field_api.

There would be two new subroutines in the API: FIELD_CLONE_ON_HOST and FIELD_CLONE_ON_DEVICE.

Documentation is updated and several tests are added.


dareg:
Some clarifications as Philippe told something was not very clear. We take into account the case where the field to be clone is optional not because we won't give the second argument to the subroutine. But because we might have a chain of subroutines being called with an optional argument being not present from the top and propagated down to the clone subroutine.

CMakeLists.txt
```diff
@@ -100,7 +100,8 @@ list(APPEND prec_srcs
     ${CMAKE_CURRENT_BINARY_DIR}/src/core/field_module.F90
     ${CMAKE_CURRENT_BINARY_DIR}/src/util/field_array_module.F90
     ${CMAKE_CURRENT_BINARY_DIR}/src/util/field_array_util_module.F90
-    ${CMAKE_CURRENT_BINARY_DIR}/src/util/field_util_module.F90)
+    ${CMAKE_CURRENT_BINARY_DIR}/src/util/field_util_module.F90
+    ${CMAKE_CURRENT_BINARY_DIR}/src/util/field_clone_module.F90)
```
awnawab, line 104: Are we sure this is a precision dependent source? I don't think it is, and we only need to compile it once and can remove it from this list.

src/util/CMakeLists.txt
```diff
@@ -38,4 +38,5 @@ field_api_add_object_library(
 set_source_files_properties( ${CMAKE_CURRENT_BINARY_DIR}/field_array_module.F90
                              ${CMAKE_CURRENT_BINARY_DIR}/field_array_util_module.F90
                              ${CMAKE_CURRENT_BINARY_DIR}/field_util_module.F90
+                             ${CMAKE_CURRENT_BINARY_DIR}/field_clone_module.F90
```
awnawab, line 41: Same as above, we only need to disable compilation for precision dependent sources.

awnawab (changes requested):
Thanks @dareg for this contribution. I just have a few minor comments that I think should be addressed before we merge this.

Just out of curiosity, so that I can understand the use case a bit better, is there a benefit to putting the FIELD on device beyond using it as a struct for `DEVPTR`?

dareg:
Hi @awnawab,
I think I've fixed the problems you've mentioned.
But for some reasons the test named test_field_clone_device.F90 is now crashing when using the serial macro, while it was working with the kernels macro.
I need to debug it (but I will be on holidays for 10 days)

```
./tests/test_field_clone_device.x 
DEV_ALLOC_SIZE = 68719476736
Failing in Thread:1
Accelerator Fatal Error: call to cuMemcpyDtoHAsync returned error 1 (CUDA_ERROR_INVALID_VALUE): Invalid value
File: /home/gmap/mrpm/grassetj/field_api_builder/field_api/build/src/util/field_1rd_clone_module.F90
 Function: field_1rd_clone_on_device:51
 Line: 79
```

awnawab:
Hi @dareg, 

Thanks for addressing the changes. That is indeed a very strange failure. In that case, please do revert it to kernels and I'll find a suitable workaround for openmp when I rebase #92 once the current PR is merged.

dareg:
Reverted to kernels

awnawab (approved):
Thanks a lot for addressing the changes, this is now G2G 👌

PR #91, opened 2025-03-25
Fix the macro call for serial

dareg:
Unless I'm wrong we cannot write this code at the moment:
```
$:offload_macros.serial()
!Fortran code goes here
$:offload_macros.end_serial()
```

We must write instead
```
$:offload_macros.launch_serial_kernel()
!Fortran code goes here
$:offload_macros.end_serial_kernel()
```

I think it's a mistake as most if not all the other macros have the same function name and macro name.

### DIFF
--- a/python_utils/offload_macros.py
+++ b/python_utils/offload_macros.py
@@ -318,7 +318,7 @@ def serial(**kwargs):
     """
 
     backend = _get_offload_backend()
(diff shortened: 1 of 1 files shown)


awnawab (approved):
Yes, great catch! Sorry about that, thanks for mopping up my mistake 🙏  Please feel free to merge once CI is green.

PR #92, opened 2025-04-03
NVHPC openmp offload backend

awnawab:
This PR adds support for NVHPC's implementation of OpenMP offload.

PR #93, opened 2025-04-07
Add `SYNC_DEVICE_WRONLY` method

awnawab:
A small PR that adds a `SYNC_DEVICE_WRONLY` method and adds the `SYNC_ON_FINAL` option to the `FIELD_ARRAY` constructor. This has no conflicts with PRs #90 or #92 and can thus be merged before them.


mlange05 (approved):
Look good to me. Many thanks.

awnawab:
Hi @pmarguinaud,

Could you please give this small PR a quick look at your earliest convenience? Once this is merged I would like to file other bigger developments. Thanks!

dareg:
Hello @awnawab,
Could you add the test-case in it's own file instead of appending it to an already existing test-case ?
Also it seems the test on the value of the 'okay' variable at the end is missing so it was not really testing much.

awnawab:
Great catch! Yes sure will do that. Now that you are back at work ( I hope you enjoyed your holiday 😄 ), could you please take another look at #90? It would be great to close that 🙏

dareg:
Hello @awnawab,
Thanks for adding the new test-case, but you forgot to remove the code for testing SYNC_DEVICE_WRONLY in the test-case of GET_DEVICE_DATA_WRONLY.
Also, could you add the new method in the documentation, at least in the public api listing.
And if you feel like it, you can even write a line in the documentation to describe what it does ! (But since none of the other sync_* methods have documentation, apart the public api listing, I won't be annoying with that)

awnawab:
Yikes, sorry about that 😅 I'll mop up that mistake and also update the Readme 👍

dareg (approved):

PR #94, opened 2025-04-07
[Partial Offload] Add Support for Blocked and Asynchronous Memory Transfers

wertysas:
This PR solves the issue [Add support for partial offloading of field API fields to device #68](https://github.com/ecmwf-ifs/field_api/issues/68) by making it possible to offload partial fields (blocks) to the device and also offload different blocks asynchronously over different streams/queues. The PR includes the features of the currently open PR [Add UNDEFINED status and force copy methods that always transfer data #85](https://github.com/ecmwf-ifs/field_api/pull/85) and adds the partial offload features through four newly introduced force methods:

* ``SUBROUTINE GET_HOST_DATA_FORCE(SELF, PPTR, QUEUE, BLK_BOUNDS, OFFSET)``
* ``SUBROUTINE SYNC_HOST_DATA_FORCE(SELF, QUEUE, BLK_BOUNDS, OFFSET)``
* ``SUBROUTINE GET_DEVICE_DATA_FORCE(SELF, PPTR, QUEUE, BLK_BOUNDS, OFFSET)``
* ``SUBROUTINE SYNC_HOST_DATA_FORCE(SELF, QUEUE, BLK_BOUNDS, OFFSET)``

Two new optional arguments are introduced in the ``FORCE``-data methods:
* ``BLK_BOUNDS`` an integer array of size 2 (``[BLK_START, BLK_END]``) that lets the user define a *block* (slice of the field in its final dimension) that will make the method only copy the block. If the field is unallocated on the device before this copy method is invoked, then the device allocation inside the copy method will only be the size of the block.
* ``OFFSET`` an integer that lets the user add an offset into the allocated device memory for offloading multiple fields into the same buffer.

These methods all set the field's status to ``UNDEFINED`` and the user has to set the status back to a *valid status* with the newly added ``SET_DEVICE_FRESH`` and ``SET_HOST_FRESH`` methods before the *normal* Field API routines can be used on the field. Additionally, the optional ``QUEUE`` argument has been removed from the normal copy methods and are now only exposed to the user through the ``FORCE`` copy methods. Previously, the copy methods could be called with the ``QUEUE`` argument and the status of the field would have been updated even if the data transfer hadn't been launched. By only allowing asynchronous copies through the ``FORCE`` methods the tracking logic of Field API will always work as expected when calling the normal copy routines and it will be impossible for a user to break the status of the fields when only using non-``FORCE`` copy  methods.

### Tests and Performance

The features in this PR have been implemented and tested in dwarf-p-cloudsc (https://github.com/ecmwf-ifs/dwarf-p-cloudsc), which has an open PR [Add Asynchronous and Blocked Field-API GPU variants #118](https://github.com/ecmwf-ifs/dwarf-p-cloudsc/pull/118) that is based on this implementation. The new cloudsc variants added can handle arbitrarily large problem sizes by using the partial offload feature and can also overlap memory transfers in both directions and compute.

The graph below shows the performance numbers in Mflop/s as reported by the cloudsc timer for the variants with varying problem sizes as measured on ECMWF's ATOS supercomputer on a node with an A100 40GB GPU. The non-blocked  **SCC** and **SCC-FIELD** variants are not able to handler problem sizes with more than 400K points, whereas the blocked variants have no such constraint.
async-bench

Below is a screenshot showing the results of a run profiled with Nsight systems. Note that there is an overlap of compute and data transfers in both directions.

cloudsc-async-overlap

### Example of Asynchronous and Blocked Offload

Below is a toy example showing how async blocked offload can be implemented, slightly modified from the new [test_force_async_acc.F90](https://github.com/wertysas/field_api/blob/1532266b7f7a38d7a2f3da6f928e0f2dcfe5ccb4/tests/test_force_async_acc.F90#L54-L80) test.

```fortran
  ! Allocate the device "BLOCK BUFFERS" of size 3*BUFFER_SIZE
  BUFFER_SIZE = 12
  BLK_COUNT=(FINAL_RANK+BUFFER_SIZE-1)/BUFFER_SIZE
  CALL F_PTR%CREATE_DEVICE_DATA(BLK_BOUNDS=[1,NQUEUES*BUFFER_SIZE])

  ! LOOP OVER THE BLOCKS AND COPY DATA TO HOST
  DO BLK_IDX = 0, BLK_COUNT-1
    BLK_START=BLK_IDX*BUFFER_SIZE+1
    BLK_END= MIN((BLK_IDX+1)*BUFFER_SIZE, FINAL_RANK)
    BLK_BOUNDS(1) = BLK_START
    BLK_BOUNDS(2) = BLK_END
    BLK_SIZE = BLK_BOUNDS(2) - BLK_BOUNDS(1) + 1
    QUEUE = MODULO(BLK_IDX, NQUEUES)+1
    OFFSET = (QUEUE-1)*BUFFER_SIZE

    CALL F_PTR%GET_DEVICE_DATA_FORCE(PTR_GPU, BLK_BOUNDS=BLK_BOUNDS, QUEUE=QUEUE, OFFSET=OFFSET)

    !$acc loop present(ptr_gpu) async(queue)
    DO K = BLK_START, BLK_END
      CALL MY_KERNEL(PTR_GPU(:,:,K))
     END DO
    !$acc end kernels

    CALL F_PTR%SYNC_HOST_FORCE(BLK_BOUNDS=BLK_BOUNDS, QUEUE=QUEUE, OFFSET=OFFSET)
  END DO

  ! Wait for work on all queues to finish
  DO QUEUE=1,NQUEUES
    CALL WAIT_FOR_ASYNC_QUEUE(QUEUE)
  END DO

  ! Set the host data to fresh and go back to using the normal api afterwards
  CALL F_PTR%FORCE_HOST_FRESH()
  ...
```


tests/init_owner.F90
```diff
@@ -36,4 +38,50 @@ PROGRAM INIT_OWNER
...
+                CALL FIELD_ABORT ("ERROR")
+        END IF
+
+        CALL FIELD_DELETE(OFORCE)
+
+        CALL FIELD_NEW(OFORCE, LBOUNDS=[10,1], UBOUNDS=[21,11])
+        CALL OFORCE%GET_DEVICE_DATA_FORCE(DEVPTR)
+
+        !$acc serial present(DEVPTR)
+        DO J=1,11
```
awnawab, line 65: The upper bound of this loop should be OMP_GET_MAX_THREADS() (you'll have to assign it to a temporary scalar though).

tests/test_force_async_acc.F90
```diff
@@ -0,0 +1,94 @@
...
+    DO J = 1,INNER_RANK
+    DO I=1,INNER_RANK
+    PTR_GPU(I,J,K) = 100 + K
+    END DO
+    END DO
+    END DO
+    !$acc end kernels
+
+    CALL F_PTR%SYNC_HOST_FORCE(BLK_BOUNDS=BLK_BOUNDS, QUEUE=QUEUE, OFFSET=OFFSET)
+    END DO
```
awnawab, line 77: The indentation of this `END DO` is off.

tests/test_get_device_data_bounds.F90
```diff
@@ -0,0 +1,151 @@
...
+    CALL FIELD_ABORT("ERROR HOST 3RD COLUMN OF PTR_CPU SHOULD NOT HAVE BEEN MODIFIED")
+  END  IF
+
+  PTR_CPU(:,1) = 38
+  PTR_CPU(:,2) = 38
+  PTR_CPU(:,3) = 39
+  CALL F_PTR%GET_DEVICE_DATA_FORCE(PTR_GPU, BLK_BOUNDS=[3,3])
+  !$acc serial, present(PTR_GPU), copy(OKAY)
+  DO I=1,128
+    IF ( PTR_GPU(I,J) /= 39 ) THEN
```
awnawab, line 78: We should explicitly index the second dim, so `PTR_GPU(I,3)`. J is only accidentally 3 here because of how Fortran terminates loops.

tests/test_get_device_data_bounds.F90
```diff
@@ -0,0 +1,151 @@
...
+      OKAY = .FALSE.
+    END IF
+    PTR_GPU(I,J) = 40
```
awnawab, line 81: Same as above.

tests/test_get_device_data_bounds.F90
```diff
@@ -0,0 +1,151 @@
...
+  !$acc end serial
+
+  IF (.NOT. OKAY) THEN
+    CALL FIELD_ABORT("ERROR DEVICE DATA NOT UPDATED BY GET_DEVICE_DATA_RDWR")
+  END  IF
+
+  CALL F_PTR%SYNC_HOST_FORCE(BLK_BOUNDS=[3,3])
+
+  DO I=1,128
+   IF ( PTR_CPU(I,J) /= 40 ) THEN
```
awnawab, line 92: And here too.

awnawab (changes requested):
Firstly, I want to take this opportunity to again say what an excellent piece of work this is 👏 👏 👌 I know how much hard work went in to this, and it really shows with the quality of the finished item 🙏 

I've left quite a few comments, but none of these are show stoppers, just a few bugs and instances where I think the implementation can be tightened up a little.

tests/test_force_async_acc.F90
```diff
@@ -0,0 +1,94 @@
...
+! (C) Copyright 2022- Meteo-France.
+!
+! This software is licensed under the terms of the Apache Licence Version 2.0
+! which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+! In applying this licence, ECMWF does not waive the privileges and immunities
+! granted to it by virtue of its status as an intergovernmental organisation
+! nor does it submit to any jurisdiction.
+
+
+PROGRAM TEST_FORCE_ASYNC
```
awnawab, line 11: Are we testing anything here that isn't covered in `get_device_data_multi_blk_async.F90`? If not, we can remove this test altogether.

python_utils/fieldType.py
```diff
@@ -27,6 +27,9 @@ def __init__ (self, **kwargs):
     self.viewRank = self.rank-1
     self.viewShape = ','.join ([':'] * (self.rank-1))
     self.lbptr = ', '.join (list (map (lambda i: "LBOUNDS(" + str (i+1) + "):", range (0, self.rank))))
+    self.lbptr_blk = ', '.join([ f"LBOUNDS({i}):" for i in range(1, self.rank)] + ["BLK_BOUNDS(1):"])
```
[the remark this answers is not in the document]
wertysas: Yes, I agree, I have removed them from the fieldType and construct/define them in place instead.

src/core/field_RANKSUFF_module.fypp
```diff
@@ -439,122 +449,251 @@ $:offload_macros.detach(ptr=['SELF%DEVPTR'], indent=6)
...
+      IF ( PRESENT(BLK_BOUNDS) ) THEN
+        CALL SELF%COPY_DATA_BLK(ND2H, BLK_BOUNDS=BLK_BOUNDS, QUEUE=QUEUE, OFFSET=OFFSET)
+      ELSE
+        CALL SELF%COPY_DATA (ND2H, QUEUE)
+      END IF
       CALL SELF%SET_STATUS (IOR (SELF%GET_STATUS (), NHSTFRESH))
     ENDIF
-    PTR (${ft.lbptr}$) => SELF%PTR (${','.join(':' for _ in range(ft.rank))}$)
+
+    CALL SELF%GET_HOST_POINTER(PTR, BLK_BOUNDS=BLK_BOUNDS)
```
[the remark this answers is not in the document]
wertysas: Thank you, me too :)

src/core/field_RANKSUFF_module.fypp
```diff
@@ -439,122 +449,251 @@ $:offload_macros.detach(ptr=['SELF%DEVPTR'], indent=6)
...
+    LOCAL_OFFSET = OFFSET
+  ELSE
+    LOCAL_OFFSET = 0
+  ENDIF
+  LB = LBOUND(SELF%PTR, ${ft.rank}$)
+
+  IF (  (LB + LOCAL_OFFSET < LBOUND(SELF%DEVPTR, ${ft.rank}$)) .OR. &
+     &  (LB + LOCAL_OFFSET + BLK_BOUNDS(2) - BLK_BOUNDS(1) > UBOUND(SELF%DEVPTR, ${ft.rank}$)) &
+     &) THEN
+    CALL FIELD_ABORT("THE SPECIFIED BLOCK IS NOT CONTAINED IN THE ALLOCATED DEVICE PTR")
```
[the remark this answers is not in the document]
wertysas: Good idea, I've added that.

src/core/field_RANKSUFF_module.fypp
```diff
@@ -422,6 +430,8 @@ $:offload_macros.detach(ptr=['SELF%DEVPTR'], indent=6)
 
   SUBROUTINE ${ftn}$_COPY_DATA (SELF, KDIR, QUEUE)
 
+  USE FIELD_ABORT_MODULE
```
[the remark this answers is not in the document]
wertysas: Yes, good catch, I have deleted it.

src/core/field_basic_module.F90
```diff
@@ -43,32 +43,49 @@ MODULE FIELD_BASIC_MODULE
...
+  PROCEDURE (FIELD_BASIC_SYNC_FORCE), DEFERRED :: SYNC_HOST_FORCE
+  PROCEDURE (FIELD_BASIC_SYNC_FORCE), DEFERRED :: SYNC_DEVICE_FORCE
   PROCEDURE (FIELD_BASIC_DELETE_DEVICE_DATA), DEFERRED :: DELETE_DEVICE_DATA
   PROCEDURE (FIELD_BASIC_CREATE_DEVICE_DATA), DEFERRED :: CREATE_DEVICE_DATA
 #ifdef WITH_FIAT
   PROCEDURE (FIELD_BASIC_CRC64),  DEFERRED :: CRC64
 #endif
   PROCEDURE :: SET_CHILDREN_DEVPTR => FIELD_BASIC_SET_CHILDREN_DEVPTR
   PROCEDURE :: SET_DEVICE_DIRTY => FIELD_BASIC_SET_DEVICE_DIRTY
+  PROCEDURE :: SET_UNALLOCATED => FIELD_BASIC_SET_UNALLOCATED
```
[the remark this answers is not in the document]
wertysas: You are right, I have deleted it.

src/core/field_RANKSUFF_module.fypp
```diff
@@ -439,122 +449,251 @@ $:offload_macros.detach(ptr=['SELF%DEVPTR'], indent=6)
...
-    LBOUNDS=LBOUND(SELF%PTR)
-    IF (.NOT. ASSOCIATED (SELF%DEVPTR)) THEN
-      CALL SELF%CREATE_DEVICE_DATA
-    ENDIF
     IF (IAND (SELF%GET_STATUS (), NDEVFRESH) == 0) THEN
-      CALL SELF%COPY_DATA (NH2D, QUEUE)
+      IF ( PRESENT(BLK_BOUNDS) ) THEN
+        CALL SELF%COPY_DATA_BLK(NH2D, BLK_BOUNDS=BLK_BOUNDS, QUEUE=QUEUE, OFFSET=OFFSET)
+      ELSE
+        CALL SELF%COPY_DATA (NH2D, QUEUE)
```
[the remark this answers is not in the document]
wertysas: I don't think we can dot that, because if a ``FORCE`` method is called without a ``BLK_BOUNDS`` argument it may still be asynchronous and in this case a call to ``COPY_DATA`` with a ``QUEUE`` argument is possible.

src/core/field_RANKSUFF_module.fypp
```diff
@@ -439,122 +449,251 @@ $:offload_macros.detach(ptr=['SELF%DEVPTR'], indent=6)
...
+  LB = LBOUND(SELF%PTR, ${ft.rank}$)
+
+  IF (  (LB + LOCAL_OFFSET < LBOUND(SELF%DEVPTR, ${ft.rank}$)) .OR. &
+     &  (LB + LOCAL_OFFSET + BLK_BOUNDS(2) - BLK_BOUNDS(1) > UBOUND(SELF%DEVPTR, ${ft.rank}$)) &
+     &) THEN
+    CALL FIELD_ABORT("THE SPECIFIED BLOCK IS NOT CONTAINED IN THE ALLOCATED DEVICE PTR")
+  ENDIF
+
+  HST_BLK => SELF%PTR(${ft.hst_blk}$)
+  DEV_BLK => SELF%DEVPTR(${':,'*(ft.rank-1)}$ LB+LOCAL_OFFSET : LB+LOCAL_OFFSET+BLK_BOUNDS(2)-BLK_BOUNDS(1))
```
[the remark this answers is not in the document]
wertysas: Yes, I have added a description of the offset in the Readme now, and an example showing how it can be used.

Readme.md
```diff
@@ -199,20 +199,70 @@ init_debug_value_jpim* to a custom value.
...
+Furthermore, the routines above add an optional dummy argument ``QUEUE`` that
+can be used to invoke asynchronous data transfers (see more below).
+
+If any data transfer routine from the advanced API has been used,
+then the user must explicitly set the status of the field before the data
+transfer methods from the default API can be used.
+To this end, there are two methods that can be used to set the internal status of the
+field:
+* ``FORCE_DEVICE_FRESH`` - should be called if the device data is up to date.
+* ``FORCE_HOST_FRESH`` - should be called if the host data is up date.
```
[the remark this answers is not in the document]
wertysas: Yes, I had forgotten that, updated it now.

tests/test_force_async_acc.F90
wertysas: No, this test is simpler than ``get_device_data_multi_blk_async``. One big difference is that there is an unstructured data region surrounding the kernels in the other test, and not in this one. If I remember correctly the other test doesn't execute asynchronously with all compiler versions and therefore I added this as well.

I have used Nsight systems to verify that this test actually executes asynchronously with several NVHPC compiler versions. But I'm happy to remove it.

src/core/field_RANKSUFF_module.fypp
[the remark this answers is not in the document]
awnawab: Ah yes, of course. Nvm, sorry for the noise.

tests/test_force_async_acc.F90
awnawab: In that case I would just add the above explanatory comment about how this test works more reliably across nvhpc compiler versions and keep it, thanks!

awnawab (approved):
Many thanks @wertysas for addressing all the requested changes 🙏 This now looks ready to merge once you add a small comment explaining how the simpler streaming test works more reliably across nvhpc compiler versions.

And once again, excellent work 👏

Readme.md
```diff
@@ -199,20 +199,76 @@ init_debug_value_jpim* to a custom value.
...
-interleave the data transfer with the computations. To do so you can add the
-QUEUE parameter, which must have a value greater than or equal to 1, when calling
-the aforementioned subroutines. With this QUEUE parameter the user will specify on
-which queue he wants the data transfer to happen, and the subroutines will return
-without waiting for the data transfer to finish. It is up to the user to be sure
-the data transfer has been done when he actually wants to use the data. This can
-be checked by using the WAIT\_FOR\_ASYNC\_QUEUE subroutine.
+## Data Transfers
+There are two categories of data transfer methods in Field API. The *core API* consists of methods that internally keeps track of the status of a field and the *advanced API* which consists of methods that relies on the user to
+keep track of where the data is located. There is no benefit in using the advanced api if the default API can be used and it is recommended to only use the features from the advanced API when
```
mlange05, line 204: `api` -> `API`

src/core/field_constants_module.F90
```diff
@@ -17,6 +17,7 @@ MODULE FIELD_CONSTANTS_MODULE
 INTEGER (KIND=JPIM), PARAMETER :: NHSTFRESH =      INT(B'00000010', KIND=JPIM)
 INTEGER (KIND=JPIM), PARAMETER :: UNALLOCATED =    INT(B'00000100', KIND=JPIM)
 INTEGER (KIND=JPIM), PARAMETER :: UNINITIALIZED =  INT(B'00000011', KIND=JPIM)
+INTEGER (KIND=JPIM), PARAMETER :: UNDEFINED =      INT(B'00001000', KIND=JPIM)
```
mlange05, line 20: Just for my own education, what makes this different to `UNINITIALIZED`? Would the in-place logic for field tracking work the same after this one was used?

mlange05 (approved):
Looks great to me! I've only left a minor cosmetic suggestion and a question, but otherwise this is GTG for me.

And just to reiterate the sentiment, this is a fantastic piece of work that shows a very neat API extension for further advanced memory transfer optimisations. The graphs tell the whole story, but I think this is very neatly implemented and should certainly be maintained and tested. Many thanks! 🙏 :shipit:

src/core/field_constants_module.F90
awnawab: `UNINITIALIZED` returns true for `IOR` checks with both `NHSTFRESH` and `NDEVFRESH`, which prevents a copy of uninitiali*s*ed data. With `UNDEFINED` the safeties are all off; you will always copy data in the requested direction regardless of the prior status.

wertysas: I think it is a very valid question and both me and @awnawab spent quite some time thinking about how to do this in a good way.

The different status values and how they trigger data movements  in the ``GET_DEVICE/HOST_DATA`` routines are summarised below. The ``UNINITIALIZED`` status never triggers any data movement because of its lower two bits being ``11``. Any constant with the lower to bits set to ``00`` will always trigger data movements, and is therefore a possible candidate for the undefined status. However, the unallocated status already has the pattern ``100`` and a field should not enter into this status if it has been allocated. Therefore the next possible value in the bit field (``1000``) is used for the ``UNDEFINED `` status.

It would be possible to modify the ``GET_DEVICE/HOST_DATA`` routines such that they don't modify the lower two bits, used to track ``HSTFRESH``and ``DEVFRESH``,   if the "undefined bit" is set. However, I didn't want to increase the complexity of these two routines by introducing two further ``IF``-checks that only are required for the force methods. In the end the user would have to "reset" the field back from the ``UNDEFINED`` state anyway, which is not very different from the way it is now, where the user must set the state explicitly with ``FORCE_HOST/DEVICE_FRESH`` after using the ``FORCE``methods.

### Field API Constants
```fortran
INTEGER (KIND=JPIM), PARAMETER :: NDEVFRESH =      INT(B'00000001', KIND=JPIM)
INTEGER (KIND=JPIM), PARAMETER :: NHSTFRESH =      INT(B'00000010', KIND=JPIM)
INTEGER (KIND=JPIM), PARAMETER :: UNALLOCATED =    INT(B'00000100', KIND=JPIM)
INTEGER (KIND=JPIM), PARAMETER :: UNINITIALIZED =  INT(B'00000011', KIND=JPIM)
INTEGER (KIND=JPIM), PARAMETER :: UNDEFINED =      INT(B'00001000', KIND=JPIM)
INTEGER (KIND=JPIM), PARAMETER :: NH2D = 1, ND2H = 2
INTEGER (KIND=JPIM), PARAMETER :: NRD = INT(B'00000001', KIND=JPIM)
INTEGER (KIND=JPIM), PARAMETER :: NWR = INT(B'00000010', KIND=JPIM)
INTEGER (KIND=JPIM), PARAMETER :: NF2L = 1, NL2F = 2
```


| FIELD STATUS     | VALUE | BIT PATTERN |
| ------------------ | ------| ------------| 
| ``NDEVFRESH``      | 1     | 00001       |
| ``NHSTFRESH``      | 2     | 00010       |
| ``UNINITIALIZED``  | 3     | 00011       |
| ``UNALLOCATED``    | 4     | 00100       |
| ``UNDEFINED``      | 8     | 01000       |


### ``GET_DEVICE_DATA``

**Transfers data if ``IAND (SELF%GET_STATUS (), NHSTFRESH) == 0``:**
  * iff 1st bit of ``STATUS`` is **not** set to 1
  * iff ``STATUS`` $\in \{0, 2, 4, 6, 8, ... \}$
  * iff ``STATUS == NHSTFREH, UNALLOCATED, UNDEFINED``

**Does not transfer data if:**
  * ``STATUS`` $\in \{1, 3, 5, 7, ... \}$
  * ``STATUS == NDEVFRESH, UNINITIALIZED``


```fortran
IF (IAND (SELF%GET_STATUS (), NDEVFRESH) == 0) THEN
  IF ( PRESENT(BLK_BOUNDS) ) THEN
    CALL SELF%COPY_DATA_BLK(NH2D, BLK_BOUNDS=BLK_BOUNDS, QUEUE=QUEUE, OFFSET=OFFSET)
  ELSE
    CALL SELF%COPY_DATA (NH2D, QUEUE)
  END IF
  CALL SELF%SET_STATUS (IOR (SELF%GET_STATUS (), NDEVFRESH))
END IF

CALL SELF%GET_DEVICE_POINTER(PTR, BLK_BOUNDS=BLK_BOUNDS, OFFSET=OFFSET)

IF (IAND (MODE, NWR) /= 0) THEN
  CALL SELF%SET_STATUS (IAND (SELF%GET_STATUS (), NOT (NHSTFRESH)))
ENDIF
```

### ``GET_HOST_DATA``
**Transfers data if ``IAND (SELF%GET_STATUS (), NHSTFRESH) == 0``:**
  * iff 2nd bit of ``STATUS`` is **not** set to 1
  * iff ``STATUS`` $\in \{0, 1, 4, 5, 8, ... \}$
  * iff ``STATUS == NDEVFRESH, UNALLOCATED, UNDEFINED``

**Does not transfer data if:**
  * ``STATUS`` $\in \{2, 3, 6, 7, ... \}$
  * ``STATUS == NHSTFRESH, UNINITIALIZED``

```fortran
IF (IAND (SELF%GET_STATUS (), NHSTFRESH) == 0) THEN
  IF ( PRESENT(BLK_BOUNDS) ) THEN
    CALL SELF%COPY_DATA_BLK(ND2H, BLK_BOUNDS=BLK_BOUNDS, QUEUE=QUEUE, OFFSET=OFFSET)
  ELSE
    CALL SELF%COPY_DATA (ND2H, QUEUE)
  END IF
  CALL SELF%SET_STATUS (IOR (SELF%GET_STATUS (), NHSTFRESH))
ENDIF

CALL SELF%GET_HOST_POINTER(PTR, BLK_BOUNDS=BLK_BOUNDS)

IF (IAND (MODE, NWR) /= 0) THEN
  CALL SELF%SET_STATUS (IAND (SELF%GET_STATUS (), NOT (NDEVFRESH)))
ENDIF
```

wertysas:
Thank you @awnawab and @mlange05 for your comments and all your help with this work!

I fixed the typo in the readme and have also added a comment here in the PR that describes the choice of the undefined logic and bit pattern. From my point of view, this is ready to be merged now.

PR #96, opened 2025-04-28
Core: Debug builds and 0-sized field fixes

mlange05:
This PR adds debug builds, which include more rigorous bounds checking, to the upstream CI workflows and fixes an initial mismatch between the shapes of the DEV and HST pointers during device offload of 0-sized fields. It then also fixes the behaviour of the FIELD and FIELD_ARRAY constructors when wrapping 0-sized arrays and adds those corner cases to existing tests.


awnawab (approved):
Many many thanks for this 🙏  I think this is the final missing piece (famous last words) for safe handling of 0-sized allocations in F-API, so it's great to complete that. Plus the debug CI tests will also help us catch problems earlier 👌 And bonus points for not creating any merge conflicts with #92 (which I am tired of rebasing 😅) 👏

dareg:
Hello @mlange05,
Could you add new test files instead of appending new testing code into the already existing test files please? I would like to keep the test as unitary as possible. Also I think it helps debugging: if there's only a single functionality being tested in a file then it's easier to know what's wrong when this unit-test doesn't complete.

mlange05:
Hi @dareg , yes fair point. I've moved the new checks into standalone tests now.

dareg (approved):

PR #98, opened 2025-05-13
Add `FIELD_STACK` abstraction

awnawab:
This PR contributes a new abstraction for a group of fields allocated in a packed buffer, called the `FIELD_STACK`. The key difference between a `FIELD_STACK` and a `FIELD_GANG` is that a gang is guaranteed to have children of the same shape and size, whereas the members of a stack can have arbitrary sizes and a shape of either rank N or N-1 (N being the rank of the stack).

The `FIELD_STACK` was designed to encapsulate a pervasive IFS design pattern, e.g. PB1, radiation interpolation, compo aerosol fields, where fields of various sizes and shapes are packed into an NPROMA blocked buffer. Moreover, such data-structures are accessed both via named members and as a buffer, and the `FIELD_STACK` allows both modes of access on device. Therefore the `FIELD_STACK` enables GPU offload of such data-structures without mandating an invasive rewrite of certain IFS data-structures. An important point to highlight here is that data movement always happens on the packed buffer, individual members cannot move independently.

Finally, the `FIELD_STACK` is targeted for use in the upcoming ECMWF benchmark.


mlange05 (changes requested):
A very nice addition that maps to various use cases in IFS code sections. My main comment at this stage is the lack of in-line and general documentation. Due to the advanced nature of the API, some helper comments on the expected layout of input maps and a general description in the README would help make this much more accessible.

I've also left a few questions on potential API extensions to make the user-side code look more elegant, and a general question on who should "own" the member-pointer map. 

But overall this is a very neat implementation of an advanced API extension that is very useful to capture more use cases in IFS. Many thanks!

awnawab:
Hi Philippe,
 
We asked for your review because we value your technical feedback on the implementation of the FIELD_STACK, and because we don't want to inadvertently add something that breaks a pre-existing use-case, or would cause merge conflicts with planned developments. It's worth noting here that the FIELD_STACK and FIELD_GANG can both be disabled at the cmake level, and the only footprint they have on the rest of the code are a few interface definitions in the factory module.
 
As a matter of policy, we can only compose our benchmark from tagged releases on main trunks, and using feature branches is not an option. As such, I will address Michael's feedback shortly, and in the absence of any concrete reason not to do so, I will merge this PR to main in one week's time.
 
Thanks again for all the feedback and discussion.

awnawab:
Thanks a lot @mlange05 for the review and feedback 🙏 The suggestion to provide a more convenient way of accessing stack members was an excellent one, as it really cuts down the boilerplate in the user-code. And thanks as well for pulling me up on the lack of documentation, this is a bad habit I really need to kick 😅 

Please have another look at the code when you can and let me know if you would like any further updates.

mlange05 (approved):
Thanks for addressing the comments. All looks good to me now. 👍 :shipit:

PR #100, opened 2025-05-21
Extension of ATOS CI. Add arch files for Intel/Gnu Atos.

piotrows:
This PR in its current form should depend on the acceptance of the HDF5-related PR. It aim is twofold:

1. It extends the CI setup to the latest Intel Classic  and the recent Gnu compiler stacks for more coverage. Gnu and Intel arch files are added to the repo
2. It enables HDF5+MPI for proper testing of the FieldAPI HDF5 I/O. Inclusion of MPI is permanent but it should have no effect without HDF5 enabled.

Fiat version is bumped to a recent 1.5.0.
There is some code duplication between NVHPC and Gnu/Intel, however, it is not clear to me if it is worth to avoid such duplication.

### DIFF
--- a/arch/ecmwf/hpc2020/gnu/11.2.0/env.sh
+++ b/arch/ecmwf/hpc2020/gnu/11.2.0/env.sh
@@ -1,0 +1,38 @@
+# (C) Copyright 1988- ECMWF.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+# Store tracing and disable (module is *way* too verbose)
+{ tracing_=${-//[^x]/}; set +x; } 2>/dev/null
+
+module_load() {
+  echo "+ module load $1"
+  module load $1
+}
+module_unload() {
+  echo "+ module unload $1"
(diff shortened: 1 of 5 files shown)


arch/ecmwf/hpc2020/gnu/11.2.0/env.sh
```diff
@@ -0,0 +1,38 @@
+# (C) Copyright 1988- ECMWF.
```
awnawab, line 1: Thanks a lot for adding the env files, these are super handy 👌 I think the toolchain files are overkill though and unnecessary. Up until now, the default ecbuild flags, plus whatever is set inside FIELD_API's compiler flags macro, has been enough to get correct behaviour on Intel and GNU. So which of these flags or ecbuild options are necessary?

If the tests pass without the toolchain files, I would be happier if both the Intel and GNU toolchains were to be removed and only the env files were retained.

awnawab (changes requested):
Many many thanks @piotrows for updating the CI to also test the I/O functionality on the ATOS 🙏  And a big apology for getting to this so late 😬 

In principle this is all fine, but I think it can be cleaned up a little. Firstly, I think the toolchain files should not be necessary, and just the env files on their own should suffice. Secondly, I've left suggestions as to how we can reduce some of the duplication in the updated build-hpc.yml workflow to make it a little nicer.

piotrows:
@awnawab I have rewritten it according to your suggestions and removed Github CI content that we have in a different PR. Please let me know how you like it now.

dareg (approved):

PR #102, opened 2025-06-06
Enable field group abstractions for Cray CPU builds

awnawab:
This PR enables field group abstractions, namely the `FIELD_GANG` and `FIELD_STACK`, for Cray CPU builds.

NVHPC, GNU and Intel (optimised) builds do not strictly enforce the `CONTIGUOUS` attribute on pointers, and will happily let you assign a non-contiguous target to a "contiguous" pointer. This is doubly paradoxical in the case of NVHPC, as it hard relies on the `CONTIGUOUS` attribute for any pointer registered in the OpenACC map; I have gotten runtime errors for 1D(!) pointers that were missing the contiguous attribute.

Amd-flang (rocm-afar) and Cray are stricter in this regard, and throw compile-time errors when trying to associate a contiguous pointer to a discontiguous target. In the case of flang, it also doesn't require the contiguous attribute for the purposes of offload, and I was able to successfully use OpenMP offload (for a meaningful ecwam test-case) with the contiguous attribute removed from `FIELD%DEVPTR`: https://github.com/awnawab/field_api/blob/797599c1969f9338281b1a5a702cae37d2ef2fe3/src/core/field_RANKSUFF_module.fypp#L38.

When we add the Cray OpenACC/OpenMP backend, we will have to carefully check whether that also requires the `CONTIGUOUS` attribute for offload. If it does, we can remove the ifdef I have added and I will come up with a different workaround for field group abstractions.


mlange05 (approved):
Looks good to me.

PR #105, opened 2025-08-01
Improvement of delayed allocation mechanism

dareg:
In the current field api version, the delayed allocation mechanism is very basic. One can ask to delay the allocation of the pointer to the CPU data and the pointer to the GPU data but the two are still allocated at the same time.
With this PR the delayed allocation mechanism is improved, and the the allocations of the CPU pointer and the GPU pointer are entirely separated. If the data is never accessed on CPU, then the CPU pointer is never allocated, same on the GPU side.

Main changes:
- delayed mechanism rewritten
- new way of managing field api status. No more get_status with bit comparison. Can now use proper add_status, remove_status, and is_status subroutines
- store lbound and ubound in wrapper field (was necessary and simplify the code since it reduce the difference between wrapper and owner)
- get_view is now callable even when the data are not on CPU or not allocated. They will be moved or allocated if necessary by get_view
- update NVHPC compiler version used in CI because of a bug in test_gang. Was 23.5, is now updated to 24.5, first version to compile it properly.


awnawab (changes requested):
Many thanks @dareg for this contribution 🙏 Most of the changes look fine to me, and some things like the more explicit `ADD_STATUS/REMOVE_STATUS` are big and very welcome improvements. I've left a few minor comments, mostly about redundant constants or code paths now being removed.

The only major disagreement I have is about the changes to `%GET_VIEW`. The abort there is an incredibly useful safety mechanism, and being forced to synchronise explicitly before calls to `%GET_VIEW` has saved us from many many painful debugging sessions. In a codebase as complex as the IFS, I think enforcing explicit synchronisation is a better, safer and simpler design choice.

As far as I understand your work, reverting the changes to `%GET_VIEW` shouldn't affect any of the new delayed allocation functionality, and you would just have to do the synchronisation explicitly before calling `%GET_VIEW`.

dareg:
Hello @awnawab, thanks for your feedbacks. I will address them in the next days.

Regarding the new get_view behaviour, Philippe would really like to keep it. So would it be fine for you if we keep the new behaviour, but make it optional? By default get_view will abort when the data are not on CPU.

awnawab:
Hi @dareg. Sure that works for me, as long as we keep the current behaviour as the default. Thanks!

src/core/field_RANKSUFF_module.fypp
```diff
@@ -306,36 +296,59 @@ CONTAINS
...
-    IF(IAND (SELF%GET_STATUS (),NHSTFRESH) /= NHSTFRESH) THEN
-      CALL FIELD_ABORT ("GET_VIEW WAS CALLED, BUT DATA IS NOT PRESENT ON HOST")
+    ${ft.type}$, POINTER :: ZPTR(${ft.shape}$)
+
+!When the first thread reach this line it checks if the data are available on cpu.
+!If not:
+! - if GET_VIEW_ABORT is TRUE then GET_VIEW will abort the program
+! - otherwise it then reachs a critical area, and check again in case another
+!   thread has moved them meanwhile. And if they are still not, then they are moved.
+    IF(.NOT. ASSOCIATED(SELF%PTR) &
```
awnawab, line 322: A wrapped 0-sized field can also return `.FALSE.` for this association test, and we can't tell such a field apart from a delayed allocation owned field. In this case we'll wrongly call `FIELD_ABORT`. We have quite a few wrappers around 0-sized fields, and we should be able to safely call `GET_VIEW` on them, as typically the block dimension is not the one with zero size.

awnawab: The cleanest workaround I can think of is to reinstate an `UNALLOCATED` status, and mark a delayed allocation field as such. Then the condition here would be `SELF%IS_STATUS(UNALLOCATED) .OR. (SELF%IS_STATUS(NDEVFRESH) .AND. .NOT. SELF%IS_STATUS(NHSTFRESH))`. All the checks in the new delayed allocation logic should then be based on the `UNALLOCATED` status rather than checking the association status of `SELF%PTR`.

awnawab: Hi @dareg,

Thanks for looking into this. Which compiler did you test this with? When I described the above issue I was testing your branch with nvhpc 24.5, it's very possible that different compilers behave differently in such a case. I was also testing within the IFS, so I hope it's not a case that the above behaviour only manifests at a certain scale.

awnawab: Thanks a lot for adding another test and for testing on our cluster. It's possible I did something wrong when I tested this earlier, let me please try your branch again in a full IFS GPU run on our side, and if I don't hit any issues then we can finalise this PR.

awnawab (approved):
Many thanks @dareg for addressing all the requested changes 🙏  I've tested this offline and I can confirm the default `GET_VIEW` behaviour is as before 👌

PR #113, opened 2025-09-11
Fix `FIELD_STACK` baseline case

awnawab:
When no `MEMBER_MAP` is specified, it is assumed that each member corresponds to 1 entry in the RANK-1 dim i.e. just like a `FIELD_GANG`. This option was broken, and the current PR fixes it. With this PR merged, we can begin the process of deprecating the `FIELD_GANG`, leaving the `FIELD_STACK` as the only abstraction for groups of fields.


dareg (approved):
Hello,

It's okay for us to mark the gang functionality as being deprecated and remove it in the future.

Issue #111, opened 2025-09-09
Compilation error with gfortran in single precision due to field_defaults_module.F90


dhaumont:
## What happened?

In field_api v0.3.2,  field_defaults_module.F90 is compiled in double precision only. This generates compilation error later on, when importing FIELD_DEFAULTS_MODULE and compiling in single precision with the gnu compiler.

### Bug fix in v0.3.2

The bug fix for `v0.3.2` is appending `field_defaults_module.F90`  to the list `prec_srcs` of files compiled in single and double precision:
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 54fff0c..8968b30 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -163,7 +163,7 @@ if( HAVE_ACC )
   ecbuild_warn("OpenACC builds force static linking.")
 endif()
 
-list(APPEND prec_srcs field_module.F90 field_array_module.F90 field_array_util_module.F90 field_util_module.F90)
+list(APPEND prec_srcs field_module.F90 field_array_module.F90 field_array_util_module.F90 field_util_module.F90 field_defaults_module.F90)
 list(REMOVE_ITEM srcs ${prec_srcs})
 
 ## build precision independent source files
```

### Bug fix  in main branch

Not tested nor investigated.
However, I had a look at the definition of field_defaults_module.F90 compilation declared in https://github.com/ecmwf-ifs/field_api/blob/main/src/core/CMakeLists.txt, and it seems to me that the bug has not been corrected.

### Reproducer 1 - Compile IAL-bundle CY50t1_deode

Compile this bundle: https://github.com/destination-earth-digital-twins/IAL-bundle/ with gfortran in single-precision. To see the bug appearing, it is necessary to remove the workaround introduced in this commit 
https://github.com/destination-earth-digital-twins/IAL/commit/e3d99aa38aa5f957b91d2eb0e42318df6d2fbdf9

```
git clone github.com:destination-earth-digital-twins/IAL-bundle.git
git checkout 481490ca3d6fb9859b4ebcf2767f75470470ef3c
./ial-bundle create
cd source/ial-source
git revert e3d99aa38aa5f957b91d2eb0e42318df6d2fbdf9
export FP_PRECISION=single
cd ../..
./ial-bundle build  --arch arch/ecmwf/hpc2020/gnu --without-double-precision  --ninja 
```

### Reproducer 2 

[...]

PR #114, opened 2025-09-12
Remove unused symbols and imports from FIELD_DEFAULTS_MODULE

awnawab:
A small bugfix PR that removes unused symbols and imports from `FIELD_DEFAULTS_MODULE` and should fix the problems reported in #111.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -170,6 +170,7 @@ foreach(prec ${precisions})
       SOURCES ${prec_srcs}
       OBJECTS
          ${obj_libs}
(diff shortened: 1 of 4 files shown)


awnawab:
Thanks @dareg for the interesting discussion. This PR fixes the issue described in #111 so we can wait to do any further cleanup at a later stage. If you are happy with it, could you please approve it? I would like to merge this unless you would like me to make some changes.

dhaumont:
Thanks @dareg and @awnawab for solving this issue. Very happy that we will be able to compile in SP with the GNU compiler

PR #117, opened 2025-09-15
Rearranged logic to require FIAT by default

piotrows:
I found the current logic behind optional FIAT dependence of Field API a bit misleading. I propose to make FIAT a required dependence, except if the path to the FIAT components is explictly specified, as it is done in CLOUDSC. Otherwise, it is suggested to the user that it is sufficient to provide the path to the three file listed to compiled FA. However, providing the path to the FIAT source doesn't work, because FIAT version of these files is way more complex that the stripped source of CLOUDSC. In conclusion, FIAT should be explicitly required by FA except if the user knows what he/she is doing and decides otherwise.

The code seems to compile in both scenarios (with FIAT and with path to the CLOUDSC-prepared sources) with NVHPC, and as a part of the CLOUDSC bundle.

### DIFF
--- a/cmake/field_api_find_fiat_modules.cmake
+++ b/cmake/field_api_find_fiat_modules.cmake
@@ -21,12 +21,12 @@
 
 macro( field_api_find_fiat_modules )
 
-   ecbuild_find_package(NAME fiat COMPONENTS ${fiat_components})
-   if( NOT fiat_FOUND )
-     if(NOT UTIL_MODULE_PATH)
-       ecbuild_critical("If not building with fiat, then the path for utility modules must be specified")
-     endif()
-   
+   if( NOT UTIL_MODULE_PATH )
+     ecbuild_info("Looking for FIAT or the path for utility modules must be specified, e.g. using -DUTIL_MODULE_PATH=")
+     ecbuild_find_package(NAME fiat COMPONENTS ${fiat_components} REQUIRED)
+   else() 
+     ecbuild_info( "UTIL_MODULE_PATH is provided. We will build independent of the full FIAT." )
+     set(fiat_FOUND 0)
      ecbuild_info( "Checking for FIAT components in ${UTIL_MODULE_PATH}" )
    
      find_file( ABOR1_PATH abor1.F90 REQUIRED
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,7 +67,7 @@ if(HAVE_IO)
     list(APPEND fiat_components MPI)
 endif()
 
-## find fiat
+## find FIAT or at least its components 
 field_api_find_fiat_modules()
 
 ## Field Gang
--- a/Readme.md
+++ b/Readme.md
@@ -17,10 +17,9 @@ Building FIELD_API requires:
 - CMake (>= 3.24)
 - [ecbuild](https://github.com/ecmwf/ecbuild) (cloned if not found)
 - [fypp](https://github.com/aradi/fypp) (cloned if not found)
-- [fiat](https://github.com/ecmwf-ifs/fiat/) (optional)
+- [fiat](https://github.com/ecmwf-ifs/fiat/) (required, may optionally be replaced with a set of prepared FIAT components.)
 
-To build FIELD_API without fiat, the path to the directory containing the utility modules `oml_mod.F90`, `abor1.F90` and `parkind1.F90` must be specified using the CMake variable `UTIL_MODULE_PATH`.
-
+To build FIELD_API without FIAT, the path to the directory containing the utility modules `oml_mod.F90`, `abor1.F90` and `parkind1.F90` must be specified using the CMake variable `UTIL_MODULE_PATH`. The files must not carry further dependencies, refer to source of CLOUDSC dwarf for a demonstration.
 ## Build and test
 ```
 mkdir build


cmake/field_api_find_fiat_modules.cmake
```diff
@@ -21,12 +21,12 @@
...
 macro( field_api_find_fiat_modules )
 
-   ecbuild_find_package(NAME fiat COMPONENTS ${fiat_components})
-   if( NOT fiat_FOUND )
-     if(NOT UTIL_MODULE_PATH)
-       ecbuild_critical("If not building with fiat, then the path for utility modules must be specified")
-     endif()
-   
+   if( NOT UTIL_MODULE_PATH )
+     ecbuild_info("Looking for FIAT or the path for utility modules must be specified, e.g. using -DUTIL_MODULE_PATH=")
```
awnawab, line 25: We can probably do without this output. Since we are invoking `ecbuild_find_package` with the `REQUIRED` keyword, ecbuild will give us a meaningful message either way.

cmake/field_api_find_fiat_modules.cmake
```diff
@@ -21,12 +21,12 @@
...
+     ecbuild_find_package(NAME fiat COMPONENTS ${fiat_components} REQUIRED)
+   else() 
+     ecbuild_info( "UTIL_MODULE_PATH is provided. We will build independent of the full FIAT." )
+     set(fiat_FOUND 0)
```
awnawab, line 29: We don't need to to initialise `fiat_FOUND` since we never use it. Also, we should replace the above info message with the original `ecbuild_critical` if `UTIL_MODULE_PATH` is not specified.

cmake/field_api_find_fiat_modules.cmake
piotrows: I disagree. The -DUTIL_MODULE_PATH is not standard, so it is helpful to have an explicit hint, at least It would have been helpful to me at the beginning.

cmake/field_api_find_fiat_modules.cmake
awnawab: If a single precision cloudsc bundle build works then we don't need `fiat_FOUND`, that would be the right test to do.

piotrows: I've checked, fiat_FOUND 0 is absolutely necessary.

awnawab: It's ok to leave it as is and manually set `fiat_FOUND` to 0 if `UTIL_MODULE_PATH` is defined. Since the build behaviour is now controlled by that variable being defined, we should name it to something more unique to avoid any crossfire e.g. `FIELD_API_UTIL_MODULE_PATH`.

A cloudsc SP build passes the compiler definition `PARKIND1_SINGLE` to field_api via `FIELD_API_DEFINITIONS`. It would be good to test if that mechanism is still working.

piotrows: > It's ok to leave it as is and manually set `fiat_FOUND` to 0 if `UTIL_MODULE_PATH` is defined. Since the build behaviour is now controlled by that variable being defined, we should rename it to something more unique to avoid any crossfire e.g. `FIELD_API_UTIL_MODULE_PATH`.
> 
> A cloudsc SP build passes the compiler definition `PARKIND1_SINGLE` to field_api via `FIELD_API_DEFINITIONS`. It would be good to test if that mechanism is still working.

I am not sure if it still holds. Grep PARKIND1_SINGLE on the latest cloudsc develop  yields:
`(field_env) napz@Zbigniews-MacBook-Pro-2 io-dwarf-p-cloudsc % grep -r PARKIND1_SINGLE *
source/field_api/CMakeLists.txt:      target_compile_definitions( ${LIBNAME}_${prec} PRIVATE PARKIND1_SINGLE )
source/field_api/tests/test_crc64.F90:#ifdef PARKIND1_SINGLE
source/field_api/tests/CMakeLists.txt:       target_compile_definitions( ${TEST_NAME}.x PRIVATE PARKIND1_SINGLE )
source/field_api/tests/CMakeLists.txt:       target_compile_definitions( ${TEST_NAME}.x PRIVATE PARKIND1_SINGLE )
source/field_api/src/core/field_module.fypp:#ifdef PARKIND1_SINGLE
source/field_api/src/util/field_util_module.fypp:#ifdef PARKIND1_SINGLE
source/field_api/src/util/field_array_module.fypp:#ifdef PARKIND1_SINGLE
source/field_api/src/util/field_array_util_module.fypp:#ifdef PARKIND1_SINGLE
source/field_api/src/io/field_hdf5_module.fypp:#ifdef PARKIND1_SINGLE
source/fiat/src/parkind/CMakeLists.txt:  target_compile_definitions( parkind_sp PRIVATE PARKIND1_SINGLE )
source/fiat/src/parkind/parkind1.F90:#ifdef PARKIND1_SINGLE
(field_env) napz@Zbigniews-MacBook-Pro-2 io-dwarf-p-cloudsc %`

Am I missing something?

awnawab: https://github.com/ecmwf-ifs/dwarf-p-cloudsc/blob/develop/src/common/module/parkind1.F90#L40

If we are passing the CLOUDSC utility modules to f-api and not building with FIAT, then f-api compiles parkind1.F90 and therefore needs the `SINGLE` preprocessor definition passed to it. This should be passed as `FIELD_API_DEFINITIONS=SINGLE` in the `--with-single-precision` cloudsc bundle option.

PR #118, opened 2025-09-19
Rearranged logic to require FIAT by default, except if explicitely demanded.

piotrows:
(Copy of #117)
I found the current logic behind optional FIAT dependence of Field API a bit misleading. I propose to make FIAT a required dependence, except if the path to the FIAT components is explictly specified, as it is done in CLOUDSC. Otherwise, it is suggested to the user that it is sufficient to provide the path to the three file listed to compiled FA. However, providing the path to the FIAT source doesn't work, because FIAT version of these files is way more complex that the stripped source of CLOUDSC. In conclusion, FIAT should be explicitly required by FA except if the user knows what he/she is doing and decides otherwise.

The code seems to compile in both scenarios (with FIAT and with path to the CLOUDSC-prepared sources) with NVHPC, and as a part of the CLOUDSC bundle.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,7 +67,7 @@ if(HAVE_IO)
     list(APPEND fiat_components MPI)
 endif()
 
-## find fiat
+## find FIAT or at least its components 
 field_api_find_fiat_modules()
 
 ## buddy allocator option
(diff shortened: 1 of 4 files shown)


awnawab (approved):
Many thanks @piotrows for addressing all the comments 🙏 If this compiles successfully with the cloudsc bundle in SP then this is G2G 👌

mlange05 (approved):
Looks good to me. I've flagged a small documentation issue, but no showstoppers.

FWIW, it seems to build fine locally with CLOUDSC (rebased over `develop-ecmwf`).

PR #120, opened 2025-10-08
Fix `%GET_VIEW` performance bug

awnawab:
As per our offline discussion last week, this PR fixes a performance bug in `%GET_VIEW`.


mlange05 (approved):
Looks good to me. 👍

PR #121, opened 2025-10-18
Increase memory pool base size and make it configurable

awnawab:
A small PR that increases the base block size of the memory pool to 3GB and makes it configurable via environment variable.


dareg:
Hello @awnawab 
Can you add some documentation about this new environment variable in the readme please?

dareg (approved):

PR #127, opened 2025-12-10
Remove defunct option to disable buddy allocator

awnawab:
The option to disable the buddy memory allocator is broken and in any case was untested. Now that we have several years of experience of running the buddy allocator across many architectures and compilers, it is safe to remove that option altogether and make the buddy allocator a hard requirement. Should this be a problem in the future it is still in the git history and will be easy to restore.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,12 +84,6 @@ endif()
 ## find FIAT or at least its components 
 field_api_find_fiat_modules()
 
-## buddy allocator option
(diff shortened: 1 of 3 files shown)


awnawab:
Hi @dareg,

Happy new year and I hope you had an enjoyable and relaxing break over the holidays. Could you please give this cleanup PR a quick look? It's a necessary precursor to a more substantial device memory allocation PR I would like to file soon.

PR #146, opened 2026-04-28
Pin ifx version

awnawab:
This PR pins the ifx version rather than always fetching the latest version, which seems to be missing ifx. Thanks for highlighting this and the fix @piotrows 🙏

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -236,7 +236,8 @@ jobs:
       run: |
         sudo apt update
         sudo apt install intel-hpckit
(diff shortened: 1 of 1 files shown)


dareg:
Hello @awnawab,
It seems the package containing the latest fortran compiler has changed. So I think the right solution is to install the correct one. See PR #147 

https://www.intel.com/content/www/us/en/developer/tools/oneapi/oneapi-toolkit-download.html?packages=oneapi-toolkit&oneapi-toolkit-os=linux&oneapi-lin=apt

PR #149, opened 2026-05-11
Make guarding of 0 sized arrays in `FIELD_ACCESS_MODULE` optional

awnawab:
This PR addresses the issue highlighted in #148 and makes the guarding of 0 sized arrays in `FIELD_ACCESS_MODULE` a compile-time option. So as to avoid excessive duplication in the CI test, I test this new option within the `DELAYED` option matrix, since these two options are orthogonal.


mlange05 (approved):
Looks good to me.

dareg (approved):

pmarguinaud (approved):

Issue #138, opened 2026-02-06
Remove `INIT_DEBUG_VALUE_JPRB` from `FIELD_DEFAULTS_MODULE`


awnawab:
### What maintenance does this project need?

Flang is stricter than most compilers regarding module imports, and in its current state we cannot do multiple precision builds using flang. To fix this, the "core" of the build, i.e. everything but the modules that explicitly do the JPRB aliasing, has to become precision independent. Therefore, `INIT_DEBUG_VALUE_JPRB` has to be removed from `FIELD_DEFAULTS_MODULE`.

PR #152, opened 2026-06-18
Another INIT_DEBUG_VALUE_JPRB hack

awnawab:
This PR contributes yet another `INIT_DEBUG_VALUE_JPRB` hack. In certain situations we can get memory corruptions because a precision dependent symbol is being linked to precision independent object libraries, which can be worked around by making it the last defined variable in the module.

field_api took a very intentional design decision to keep JPRB only as an aliased symbol, that's why `INIT_DEBUG_VALUE_JPRB` is unused throughout the code. It would be great if we could remove it and clean up all related workarounds.

This also causes other issues with flang, as outlined in #138.


mlange05 (approved):
Looks good to me. 👍

dareg (approved):

PR #154, opened 2026-06-30
Add 3D gather/scatter capability

pmarguinaud:
### Description

Both `FIELD_GATHSCAT` (already in the Field API library) and `FIELD_GATHSCAT_COLLAPSE2` (introduced in this PR) implement high-level **gather-scatter** patterns:

1. **Filter** a multi-dimensional field using a logical condition (the *trigger*).
2. **Gather** the data that satisfies the condition into a compact, contiguous array.
3. Let the user modify that compact array (e.g., on CPU or GPU).
4. **Scatter** the modified values back to their original locations in the source field.

This pattern is essential for algorithms that need to process only a subset of grid points (e.g., convective columns, cloud-covered areas, or land points); this has already proven to improve performance quite a lot.  `FIELD_GATHSCAT` is currently used in the shallow convection algorithm of ARPEGE.

`FIELD_GATHSCAT` is used when the selection condition is **2-D** and you want to keep **all intermediate dimensions** (e.g., vertical levels, variables) intact in the gathered array. It works with fields of any rank ≥ 2.

The purpose of `FIELD_GATHSCAT_COLLAPSE2` is to be used when the selection condition is **3-D**  and you need to collapse the two leading spatial dimensions of arrays into one. This is typical for physics parametrizations that operate on a subset of atmospheric cells. We will use it in the AROME micro-physics.

Several tests have been added to exercise the newly created code.

Tested on A100 with nvhpc/26.5 and MI300X with afar/23.2.0. 

Tested in AROME, with nvhpc/26.5 on A100.

### Contributor Declaration

By opening this pull request, I affirm the following:

* All authors agree to the [Contributor License Agreement](https://github.com/ecmwf/codex/blob/main/Legal/contributor_license_agreement.md).
* The code follows the project's coding standards.
* I have performed self-review and added comments where needed.
* I have added or updated tests to verify that my changes are effective and functional.
* I have run all existing tests and confirmed they pass.


tests/gather_scatter_collapse2.F90
```diff
@@ -0,0 +1,155 @@
...
+ELSE
+  ID_FILTERED => GATHER_DEVICE_DATA_RDWR (YL_FGS, YLFD)
+ENDIF
+
+! update filtered data
+
+#ifdef OMPGPU
+!$OMP TARGET
+#else
+!$ACC SERIAL PRESENT (ID_FILTERED) COPYIN (YL_FGS) IF (JPASS == 2)
```
awnawab: Compilation is failing for this test in the OpenACC builds. That would be nice to fix before this can be merged. My best guess would be a missing `CONTIGUOUS` attribute on `ID_FILTERED`.

tests/gather_scatter_collapse2_lastdim.F90
```diff
@@ -0,0 +1,147 @@
...
+    IF (SIZE (FILTERED_D, 1) /= NPROMA) CALL FIELD_ABORT ('NPROMA MISMATCH')
+    IF (LBOUND (FILTERED_D, 2) /= 1) CALL FIELD_ABORT ('LBOUND MISMATCH')
+    IF (UBOUND (FILTERED_D, 2) /= NFLD) CALL FIELD_ABORT ('UBOUND MISMATCH')
+  ENDIF
+  
+  !ACTUALLY UPDATE THE D ARRAY WITH THE MODIFIED DATA
+  
+  CALL FGS%SCATTER ()
+  
+  DO JBLK = JBLKL, JBLKU
```
awnawab, line 101: We're failing somewhere here in the cray builds. Since we don't use the gather scatter functionality, I would be ok with this test being excluded from the cray CI run (with something like a conditional `ctest -E collapse2_lastdim`.

src/shuffle/field_RANKSUFF_gather_collapseCOLLAPSE_module.fypp
```diff
@@ -0,0 +1,86 @@
+#! (C) Copyright 2022- ECMWF.
+#! (C) Copyright 2022- Meteo-France.
+#!
+#! This software is licensed under the terms of the Apache Licence Version 2.0
+#! which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#! In applying this licence, ECMWF does not waive the privileges and immunities
+#! granted to it by virtue of its status as an intergovernmental organisation
+#! nor does it submit to any jurisdiction.
+
+MODULE FIELD_${RANK}$${SUFF}$_GATHER_COLLAPSE${COLLAPSE}$_MODULE
```
awnawab: Do we really need a separate module per derived type here? If possible, I would prefer to keep the fypp expansion to the rank-suff matrix, and a loop in here for the different values of `COLLAPSE`. This is effectively what we are already doing in many places for `WRAPPER, OWNER` types in the same module. You could then simply add the relevant files to the `ranksuff_srcs` and `non_ranksuff_srcs` lists.

awnawab (changes requested):
Thanks a lot for this development @pmarguinaud 🙏 We don't use the gather-scatter functionality, so we would support replacing the existing gather-scatter with gather-scatter-collapse1. Please proceed however you see fit.

I've left a few comments otherwise. I would prefer to restrict the fypp mapping to the rank-suff matrix, and for further templating moving the loop inside the files themselves. Since this is purely a stylistic consideration, I'll leave the final call up to you. But it would be nice to have a green CI before we merge.

tests/gather_scatter_collapse2.F90
pmarguinaud: What version of nvhpc do you use ?

awnawab: 24.5 in the hpc ci.

pmarguinaud: It is quite old. Maybe we should upgrade to a 2026 release ?

I have tried 24.5 on ECMWF cluster, but I cannot reproduce the problem:

ec:/perm/sor/field_api/build

awnawab: Yes I agree, but a lot of our downstream CI also uses nvhpc 24.5 and that's not trivial to update (unlike here), so I'd like to keep the hpc ci at 24.5 (we are testing a newer 25.9 in the github CI already) until our downstream CI has been updated.

I'm happy to checkout your branch and try and reproduce the problem myself, but I'm moving house over the next week so I'll probably try this on Monday the 27th (I hope that's ok, sorry for the inconvenience).

pmarguinaud: Thank you, I will be waiting.

awnawab: I managed to reproduce the issue offline, and it's a compiler bug related to the `!$ACC SERIAL IF` directive. The cleanest workaround to me seems to be to move the condition to Fortran rather than in the openacc pragma. If you agree I can push a commit here with the workaround.

pmarguinaud: Yes I agree, please proceed

src/shuffle/CMakeLists.txt
```diff
@@ -28,6 +28,44 @@ field_api_expand_fypp(
    OUTPUT_SRCS            srcs
 )
 
+## expand collapse2 templates
+foreach (SUFF IN ITEMS IM RM RD LM)
+  string (TOLOWER ${SUFF} suff)
+  foreach (RANK RANGE 1 5)
+    foreach (FUNC IN ITEMS _shuffle _gather)
+      add_custom_command (
```
awnawab, line 36: I'm pretty sure we don't need this fypp expansion here. If we add `_shuffle_collapse2` and `_gather_collapse2`, we can delete this entire fypp expansion section.

src/shuffle/CMakeLists.txt
```diff
@@ -28,6 +28,44 @@ field_api_expand_fypp(
...
+          ${CMAKE_CURRENT_SOURCE_DIR}/field_RANKSUFF${FUNC}_collapse2_module.fypp
+          > ${CMAKE_CURRENT_BINARY_DIR}/field_${RANK}${suff}${FUNC}_collapse2_module.F90
+        DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/field_RANKSUFF${FUNC}_collapse2_module.fypp
+        VERBATIM)
+      list(APPEND srcs ${CMAKE_CURRENT_BINARY_DIR}/field_${RANK}${suff}${FUNC}_collapse2_module.F90)
+    endforeach ()
+  endforeach ()
+endforeach ()
+
+add_custom_command (
```
awnawab, line 49: Same here, adding `field_gathscat_collapse2_module field_gathscar_collapse2_type_module` to non_ranksuff_srcs should be enough and the manual fypp expansion here can be removed.

awnawab (changes requested):
Thanks a lot @pmarguinaud for incorporating my earlier suggestions 🙏 Some further small changes will really help to remove a lot of redundant cmake, so it would be nice to incorporate those.

As for the failing tests, the cce seg faults appear genuine. Considering we don't rely on this functionality, I would be happy with simply marking the new `*_lastdim` test as expected to fail for cray builds, and then you can investigate and fix it independently from this PR (I will create a github issue to track it).

The failing rocm afar tests are related to the baseline main moving since this PR was opened (apologies, but that was an important development 😬). For PRs from forks, github uses the state of the build-hpc workflow as in main, but that requires code changes you are missing here. Another merge/rebase with main should fix the rocm afar tests, considering they were green before PR #157 was merged to main.

src/shuffle/field_RANKSUFF_gather_collapseCOLLAPSE_module.fypp
pmarguinaud: I did the changes; see whether it suits you.

src/shuffle/CMakeLists.txt
pmarguinaud: I did the changes; see whether it suits you.

awnawab (approved):
Thanks a lot @pmarguinaud for cleaning up the cmake, this is now G2G ✅

PR #157, opened 2026-07-27
Enable dual precision flang builds

awnawab:
This PR makes the field_api object libraries truly precision independent, thereby enabling dual precision builds for flang, as described in #138.


dareg (approved):

mlange05 (approved):
Looks good to me.

